PR #20, opened 2025-08-04
Jakub Morawiec - zadanie 0

jmorawiec:
Próbowałem eksperymentować z type: module, ale skompilowałem to też ze starym, dobrym const require

### DIFF
--- a/0. Wstęp/intex.ts
+++ b/0. Wstęp/intex.ts
@@ -1,0 +1,14 @@
+import fs from 'fs';
+
+function eloZelo(times: number): void {
+    for (let i: number = 0; i <= times; i++) {
+            fs.appendFile('elo-żelo.txt', 'Elo żelo\n', 'utf8', (err) => {
+                if (err) console.error(err);
+            })
+    }
+}
+
+const currDate: Date = new Date()
+let currMinutes: number = currDate.getMinutes()
+
+eloZelo(currMinutes)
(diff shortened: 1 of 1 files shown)


0. Wstęp/intex.ts
```diff
@@ -0,0 +1,14 @@
+import fs from 'fs';
+
+function eloZelo(times: number): void {
+    for (let i: number = 0; i <= times; i++) {
+            fs.appendFile('elo-żelo.txt', 'Elo żelo\n', 'utf8', (err) => {
```
Leadman5555, line 5: We are in synchronous context here. As such we should use the synchronous variants of the functions. While `appendFile` may seem like a blocking call at a first glance, it uses a callback pattern. Anything that takes a callback is, overall, most probably not synchronous. While this function will work correctly thanks to file locks, it's not the correct way of doing it. You should either mark the function as async and use awaits or, preferably in this case, use the `appendFileSync` method in a try/catch block

Leadman5555 (changes requested):
Let's change the function to use either fully async methods or their synchronous counterparts. I'm also not sure what you meant about compiling with require, as you are not using it in your code - and that's good, because it's not the typescript way of handling imports.

jmorawiecpwr:
Changes have been made, please take a look @Leadman5555

PR #22, opened 2025-08-04
Hubert drewlo 0

Drewienkopl:
Wykonanie zadania 0, sam plik index.ts znajduje się w katalogu src.


Leadman5555 (changes requested):
Before I even start reviewing your work. You have forgotten to include a proper `.gitignore`  file in your code. As a result, you have pushed just over 56 thousand worth of lines of node modules. Please include a gitignore file that will exclude the node modules and other unnecessary files and folders (like the /dist folder) 

image

PR #23, opened 2025-08-04
Zadanie 0

BKmil:
Odpowiedź do zadania 0


index.ts
```diff
@@ -0,0 +1,11 @@
+import * as fs from "fs";
```
czaja307, line 1: it's a good practice to only import the methods that you're using - especially if you are using only one or few of them
```suggestion
import { writeFileSync } from "fs"
```

PR #34, opened 2025-08-05
Zadanie 0

fidok15:
Sorry za delay

### DIFF
--- a/src/intex.ts
+++ b/src/intex.ts
@@ -1,0 +1,10 @@
+import { writeFileSync } from 'fs';
+
+function eloZelo(amount: number): void {
+  const content: string = "Elo żelo\n".repeat(amount);
+  const filePath: string = "./elo-żelo.txt";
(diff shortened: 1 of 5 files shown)


src/intex.ts
```diff
@@ -0,0 +1,10 @@
+import { writeFileSync } from 'fs';
+
+function eloZelo(amount: number): void {
+  const content: string = "Elo żelo\n".repeat(amount);
+  const filePath: string = "./elo-żelo.txt";
```
Leadman5555, line 5: Using relative paths in such a way is generally not the best idea. If you want to explicitly choose the working directory, it's better to create the path using the `path` module and get current directory from, for example, `__dirname`. Nevertheless, it's just a suggestion and not a mistake.

Leadman5555 (approved):
Ala is currently not available, so I'll take over for the time being.  
One minor suggestion, but not a mistake. Feel free to either change it or to simply merge without any modifications, your call.

PR #35, opened 2025-08-05
Zadanie 0

wsparcie:
Rozwiązanie do zadania 0 - Elo Żelo.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -6,11 +6,10 @@ function eloZelo(repeats: number): void {
   writeFileSync(file, '')
   for (let i: number = 0; i < repeats; i++) {
     writeFileSync(file, text + '\n', { flag: 'a' })
-    console.log(text)
+    console.log(`${i + 1}: ${text}`)
   }
 }
 
 let minutes: number = new Date().getMinutes()
-let message: string = `minutes: ${minutes}`
-console.log(message)
+console.log(`minutes: ${minutes}`)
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -0,0 +1,15 @@
+import { writeFileSync } from 'fs'
+
+function eloZelo(repeats: number): void {
+  const text: string = 'Elo żelo'
+  const file: string = './elo-żelo.txt'
+  writeFileSync(file, '')
+  for (let i: number = 0; i < repeats; i++) {
+    writeFileSync(file, text + '\n', { flag: 'a' })
```
czaja307, line 8: disc operations (e.g. fileWrite) are quite time-consuming - try to prepare the string first and the write the file using only one `writeFileSync` call

index.ts
```diff
@@ -0,0 +1,15 @@
...
+    console.log(`${i + 1}: ${text}`)
+  }
+}
+
+let minutes: number = new Date().getMinutes()
+console.log(`minutes: ${minutes}`)
```
czaja307, line 14: console logs are a great tool for debugging, but try to avoid leaving them in the final code - they slow down the execution, but what's more important they might leak some sensitive informations

czaja307 (changes requested):
just few minor issues, but you got the idea right :D

czaja307 (approved):
🔥

PR #46, opened 2025-08-05
feat: Kamil Ramocki task 1

kamilramocki:
I'm hopeful I've .gitignored everything I should've🙏🙏🙏


czaja307 (changes requested):
pls fix CI errors (prettier, linter and test are currently failing)
if you have any difficulties during refactor, feel free to ask for help in the #techforum on discord server

czaja307 (changes requested):
Pls do not disable the SolvroConfig - I think that the issue was caused because you set the base branch to main (which had some accidental changes then) instead of your `kamil-ramocki` branch. Try to push the revert to the prettier and we'll see then

PR #47, opened 2025-08-05
Task 1 done

wdomini4801:
Created endpoint with specific path that returns JSON with given format and status code.


GOLDER303 (approved):
Wszystko git, super robota

PR #59, opened 2025-08-08
feat: Zadanie 1

ptakj:


Leadman5555 (changes requested):
CI did not pass. Probably linter errors with types. You can check the details using the "Actions" tab. In future, run the ci locally first before you create the pull request to catch any errors early on.

PR #78, opened 2025-08-10
Iga wolanin zad 0

igawolanin:
Chyba wszystko dobrze ale nie umiem w gita, może cos namieszałam przez przypadek


src/index.ts
```diff
@@ -0,0 +1,9 @@
+import * as fs from "node:fs";
```
Leadman5555, line 1: Note for future: `import *` means we import everything from a module - it's usually not needed. Here, for example, we only need a single function from `node:fs`, so it would be to import the `writeFileSync` function only with `import { writeFileSync } from "fs";`.

Leadman5555 (approved):
LGTM 👍 . You can merge it

src/index.ts
Leadman5555: With future eslint config, such imports (with `*`) will most probably be treated as errors by the linter and will not pass the ci

PR #84, opened 2025-08-11
Iga wolanin 1

igawolanin:
Solution to task 1

### DIFF
--- a/src/app.controller.spec.ts
+++ b/src/app.controller.spec.ts
@@ -16,9 +16,12 @@ describe("AppController", () => {
     appController = app.get<AppController>(AppController);
   });
 
-  describe("root", () => {
-    it('should return {"quote": "Jestem czajnikiem", "title": "Wakacyjne Wyzwanie Solvro!!!"}', () => {
-      expect(appController.getHello()).toStrictEqual(appController.getHello());
+  describe("AppController root endpoint", () => {
+    it("appController.getHello()", () => {
+      expect(appController.getHello()).toStrictEqual({
+        title: "Wakacyjne Wyzwanie Solvro!!!",
+        quote: "Jestem czajnikiem",
+      });
     });
   });
 });
--- a/src/app.controller.ts
+++ b/src/app.controller.ts
@@ -2,13 +2,18 @@ import { Controller, Get, HttpCode } from "@nestjs/common";
 
 import { AppService } from "./app.service";
 
+export interface Result {
+  title: string;
+  quote: string;
+}
+
 @Controller("wakacyjne")
 export class AppController {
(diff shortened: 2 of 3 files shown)


src/app.controller.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    const app: TestingModule = await Test.createTestingModule({
+      controllers: [AppController],
+      providers: [AppService],
+    }).compile();
+
+    appController = app.get<AppController>(AppController);
+  });
+
+  describe("root", () => {
+    it('should return {"quote": "Jestem czajnikiem", "title": "Wakacyjne Wyzwanie Solvro!!!"}', () => {
```
Leadman5555: Tests are covered in later modules. But since you decided to write one, I'm compelled to point out the mistakes.  
Different frameworks and programming languages have different ways of writing tests. Usually, you have a single test method per a functionality you want to test, sometimes divided in successful and failure results. 
In Node, the most common convention is using the describe -> it syntax. "describe" starts a test group - it details what do all test methods inside the group will do, for example 'appController' as above. 'root' is not the best name for a test group.  
The bigger problem however, is the test case itself. "it" clause contains two things:
- method description (here should be the endpoint method name for example). It never should be the full result.
- the test body itself in a function for lazy calls. Now, the test must, as the name suggests, test things. In your test case you compare the result to... the result itself. The comparison will never fail, because you are comparing two equal objects. What you should do instead, is to compare the result of `appController.getHello()` to the desired JSON value.

src/app.controller.ts
```diff
@@ -0,0 +1,14 @@
+import { Controller, Get, HttpCode } from "@nestjs/common";
+
+import { AppService } from "./app.service";
+
+@Controller("wakacyjne")
+export class AppController {
+  constructor(private readonly appService: AppService) {}
+
+  @Get("wyzwanie")
```
Leadman5555: The endpoint URI was supposed to be different, but it doesn't really matter. At least in my group.

src/app.service.ts
```diff
@@ -0,0 +1,11 @@
+import { Injectable } from "@nestjs/common";
+
+@Injectable()
+export class AppService {
+  getHello(): { title: string; quote: string } {
```
Leadman5555: You tried to specify the return type explicably - that's good. However, writing the JSON output as type is not the best approach. What I would recommend instead, is to create a custom type (or interface) with the result structure. For example: 
```typescript
interface QuoteResult { 
title: string; 
quote: string 
}
```

Leadman5555 (changes requested):
Two minor nitpicks to address. Everything else is fine 👍

PR #86, opened 2025-08-11
Jakub Morawiec - zadanie 1

jmorawiec:
Wybacz za lekkie opóźnienie. W poleceniu znalazłem drobną nieścisłość - w pliku .md kod statusu miał być ustawiony na 418, a w docsach na dysku - 200. To tak czy siak kwestia cyferki, ale mówię na zaś, że wybrałem losowo jeden z tych dwóch.

### DIFF
--- a/src/app.service.ts
+++ b/src/app.service.ts
@@ -1,0 +1,12 @@
+import { Injectable } from "@nestjs/common";
+
+@Injectable()
+export class AppService {
+  getHello(): object {
+    return {
+      title: "Wakacyjne Wyzwanie Solvro!!!",
+      quote:
+        "Unikwanie wyzwań sprawia, że stoimy w miejscu. Chyba, że chodzi o walkę z Solvro Configiem",
+    };
+  }
+}
--- a/test/app.e2e-spec.ts
+++ b/test/app.e2e-spec.ts
@@ -1,0 +1,32 @@
+import request from "supertest";
+import type { App } from "supertest/types";
+
+import type { INestApplication } from "@nestjs/common";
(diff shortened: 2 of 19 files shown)


test/app.e2e-spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  it("/ (GET)", () => {
+    return request(app.getHttpServer())
```
Leadman5555, line 23: Tests are covered in a later module, so writing them is not required as of now. Nevertheless, it's never a bad call to do so

src/app.service.ts
```diff
@@ -0,0 +1,12 @@
+import { Injectable } from "@nestjs/common";
+
+@Injectable()
+export class AppService {
+  getHello(): object {
```
Leadman5555, line 5: When returning something without a clearly defined structure, it's usually best to let the linter infer the type instead of manually specifying it - object, any and unknown are rarely useful if manually specified unless the purpose is to get rid of  linter errors. Here, instead of typing the return value as an `object`, it would be better to create a custom interface (or type):
```typescript
export interface HelloQuote {
 title: string;
 quote: string;
}
 ```
 So the best approach is to either let the linter infer the type on its own, or to create a custom one instead of using `object`

Leadman5555 (approved):
LGTM 👍 . One minor suggestion, nothing wrong. The quote is a nice addition. You can merge it.

PR #90, opened 2025-08-11
feat: jakub ptak zadanie 1

ptakj:
prettier


src/app.service.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@Injectable()
+export class AppService {
+  getHello(): JSON {
+    const returnValue: unknown = {
+      title: "Wakacyjne Wyzwanie Solvro!!!",
+      quote:
+        "Mieszany dla najsptyrtniejszych zawodników, bo są trzy w cenie jegnego",
+    };
+    return JSON.stringify(returnValue) as unknown as JSON;
```
Leadman5555: This is not correct for several reasons.  
1. First of all, the return type here is not a JSON, it's a string (`JSON.stringify()` returns a string). You cannot cast it to JSON (`as JSON`) because the types do not overlap in the slightest. `string` is just text, while `JSON` is a class with methods. Using the `as unknown` chain does nothing apart from silencing the linter and should not be used unless necessary. How do we fix? Lets not convert to JSON altogether and return the type itself.
2. The type of `returnValue` is not unknown - as the name suggests, "unknown" means you do not know the type at runtime. You **do** know it here. Instead, create a custom interface (or type) that reflects the object structure (two string fields of `title` and `quote`) and use it here.
3. This is a bit harder to explain, but we should not convert the types to string JSON ourselves, apart from few special cases. The conversion should handled by the framework. Return the custom interface here, without converting it to string. The same applies to the controller. You can even omit the type if possible and let it be inferred from usage - it will work as long as you use a custom type when declaring the `returnValue`

Leadman5555 (changes requested):
One important fix in the service. If I didn't explain it clearly, feel free to reply to the pending comment 👍

Leadman5555 (commented):
Well, you just removed the types altogether which is not really what I meant. But it's not incorrect either, so I'll let it pass. You can merge it if you want to.

Leadman5555:
I also think you have misunderstood my discord message. I wrote that the **title** of the PR should match the required format. Not the name of each commit. You are naming commits incorrectly. PR name should be `feat: jakub ptak zadanie 1` and the commit names should be describing what the commit does - `fix: service types` for example. I'll change the PR title for you this time, but remember it in the future. PR title != commit name

Leadman5555:
image

TLDR: this is wrong. You shouldn't name commits like this

Leadman5555 (approved):

Leadman5555:
I've just realized you are trying to merge into main branch again. Change the target to your branch

PR #96, opened 2025-08-13
Jakub Bylicki - task 1

MrBHP:


Leadman5555 (changes requested):
The CI did not pass. Excluding the commit names, the are several type-related errors. As per my discord message on the info channel, I will not be checking the PR until the CI passes. To stop the tests from failing, either fix them yourself, delete them altogether, or copy the solution shared on discord.

PR #102, opened 2025-08-17
feat: finished task 2

Drewienkopl:
Merged task 1 into main hubert-drewlo branch, fixed loop and stuff.

I had a test on monday last week and right after that I went to work to germany for a week and only came back on friday. I hope it still counts if I make those fixes now. Tasks 3 and 4 were done on time, it's just that for some reason, as I was doing them, I didn't think to first make fixes to this one.

FixedDiagram

Trip 1 to Many Participant (many participants for one trip).
Participant 1 to Many Expense (many expences for one participant).
ProofOfWorking2


Leadman5555 (changes requested):
Before I start checking any of the code, we need to fix the relationships diagram. There are two problems, one small, one critical.

Let's start with the small one: `spentAt` field should be optional (nullable). It doesn't really make sense otherwise. Maybe rename it too, to something like `paidAt`? Then you could create expense entries before the trip and register if and when a participant paid for their expenses. It's up to you, but the current design is a bit too rigid.  

And now, moving on to the big problem - we have a relationship loop. Imagine the following scenario:
 - Trip with Id 1, Trip with Id 2
 - Participant with Id 1 and tripId = 2
 - Expense with Id 1, tripId = 1, participantId = 1
 
We have the Expense pointing to a different Trip than the one the participant is on. Yes, we can write our persistence layer code in such a way to avoid this, but we shouldn't have to - the database itself should forbid such behavior. How to fix this? There are two solutions:
- add constraints (quick but not the best longterm)
- remodel the relationships (we choose this one here)

So, your goal is: remodel the relations between entities so that there is no "loop". I'd change the expenses table. 

Apart from that, please watch the commit names - otherwise the CI will error out. It's too late to fix it now, but keep it in mind for the future. Also, change the PR title to adhere to the given format.

I see that you've had some problem with the tests. In my opinion, writing the tests in a wrong way just so they pass is pointless. Before we reach the test module, just delete the test files - no need to bother with them 👍

Leadman5555:
Also, merge your first branch (from task 1) into your main branch (`hubert-drewlo`) before making any fixes

Leadman5555 (approved):
Looks okay at the first glance. I won't be checking it in detail as, well, we are quite behind the schedule. So I'll give you the benefit of the doubt and merge it

PR #103, opened 2025-08-17
Dawid Kowal - Etap 2, Bazy Danych

GTR1701:
This pull request sets up the initial foundation for a NestJS backend project for managing trips, participants, and expenses. It includes a database schema and migrations, seed data, and the basic application structure with modules, services, and DTOs.

**Database Schema and Seed Data**

* Defined Prisma schema for trips, participants, and expenses with enums for status and category, including relations and constraints.
* Added initial migration SQL and migration lock for PostgreSQL, establishing tables and foreign keys. [[1]](diffhunk://#diff-de68709e74d7ce6f893b999908ffcba05ce30295d45b1c11c8e656055b796769R1-R57) [[2]](diffhunk://#diff-12c83dddac45057e638a36a8d6617c917f93cd0cc8ac5d8cdc121573007db546R1-R3)
* Added `prisma/seed.ts` to populate the database with sample trips, participants, and expenses for development and testing.

**Application Structure and Features**

* Implemented main application module (`src/app.module.ts`), controller, and service, with a basic root endpoint. [[1]](diffhunk://#diff-089f4f2474b64391c42b6e66aed33977e132058d92108f0a63234a7862e1f8b8R1-R13) [[2]](diffhunk://#diff-ef0ff0ca9442bda67ad0810654d636991173ecd7c1abcd6da09288b8815759ffR1-R13) [[3]](diffhunk://#diff-a4ac7efdbcfcdcc01a90d433ec3dabf1d88ab4bffbe3dc88f13db54db2e051f5R1-R8) [[4]](diffhunk://#diff-bbc8b4e9f8b329f999b013aaedf36a1db25e79e7020f73125dd7d3e75ae27161R1-R24)
* Added Trips module with structure for managing trips, participants, and expenses, including DTOs and services. [[1]](diffhunk://#diff-96dec30b49c5743d45618083f3391075d2efcdfe87b3ab57528843395dc8242cR1-R10) [[2]](diffhunk://#diff-76a9f9329abb7532c73fbcda6c80cf3e4d70942de61f09661dfebfa33e352eb5R1-R54)

**Documentation**

* Enhanced `README.md` with API endpoint descriptions, example requests, instructions for running and seeding the app, and an overview of project structure and best practices.


GTR1701:
Diagram ERD:
Budżetownik ERD
category-enum
status-enum

src/participants/participants.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    return this.prisma.participant.create({
+      data: {
+        name: createParticipantDto.name,
+        email: createParticipantDto.email,
+        phone: createParticipantDto.phone,
+        isOrganizer: createParticipantDto.isOrganizer ?? false,
+        tripId,
+      },
```
AlicjaBonar, line 30: It should be enough to just write:
return this.prisma.participant.create({
  data: createParticipantDto,
});
However, if you do it this way, you need to handle the organizer logic first.

prisma/seed.ts
```diff
@@ -0,0 +1,160 @@
...
+    data: {
+      name: "Wakacje w Hiszpanii",
+      description:
+        "Wspaniały tygodniowy pobyt na Costa del Sol z wizytą w Madrycie",
+      status: TripStatus.PLANNED,
+      startDate: new Date("2025-07-15"),
+      endDate: new Date("2025-07-22"),
+      budget: 5000,
+    },
+  });
```
AlicjaBonar: Your solution with create works correctly, but keep in mind it may fail if you run the seed multiple times and the records already exist. To avoid this, in the lecture it was shown to use the upsert command, which is usually preferred for seeding.

src/trips/trips.service.ts
```diff
@@ -0,0 +1,227 @@
...
+      where: { id },
+    });
+
+    if (existingTrip === null) {
+      throw new NotFoundException(`Trip with ID ${String(id)} not found`);
+    }
+
+    await this.prisma.trip.delete({
+      where: { id },
+    });
```
AlicjaBonar, line 127: It would be enough to write: 
return this.prisma.trip.delete(id);

AlicjaBonar (approved):
I can see that you’ve put a lot of time into this, and it really shows! Great documentation and excellent use of DTOs. You went a bit overboard with the controllers and services, so next week’s homework will be much shorter, as it will focus mainly on that. I’ve also included one pointer that might help you avoid a problem in the future, along with a few minor cosmetic suggestions.
One more thing: this was probably intentional, but the way the models are set up makes it impossible for a single participant to join more than one trip. You could create a separate model for participants and another one to track which participants are in which trips. You don’t need to change it now, but it’s something worth keeping in mind for the future.

prisma/seed.ts
GTR1701: Right, forgot about the upserting. But would actually a full db wipe before inserts be a good idea for the seeder?

src/trips/trips.service.ts
GTR1701: Actually didn't know you could do that. Guess I've been writing unnecessary boilerplate for 3 years now xD

src/participants/participants.service.ts
GTR1701: Yeah, I was torn between the two options, but decided to go with the approach you see there.

prisma/seed.ts
AlicjaBonar: Yeah, full wipe could work too, but then you’d need to make sure the wipe always runs first — otherwise the create calls might fail on duplicates. With upsert you don’t have to worry about that, it’s safer by default even if you want to do the full wipe.

src/trips/trips.service.ts
AlicjaBonar: That's a lot of time 🙈
Happy to help :)

GTR1701: Actually, prisma's yelling at me that I can't do it and it requires the where clause
image

PR #113, opened 2025-08-17
feat/ Dominik Kaczmarek lista 2

feen1e:
ERD: 
budzetownik_diagram_2025-08-17T11_10_02 545Z

Dodałem też 2 endpointy:
Screenshot 2025-08-17 191305
Screenshot 2025-08-17 191329

Pozwoliłem sobie wyłączyć tymczasowo testy e2e w CI, bo zawsze kończyłyby się niepowodzeniem ze względu na brak `.env` w repo (lokalnie wszystkie przechodzą). Jeśli nie powinienem był tego robić to włączę z powrotem


prisma/seed.ts
```diff
@@ -0,0 +1,72 @@
...
+
+async function main() {
+  const trip = await prisma.trip.create({
+    data: {
+      destination: "Rejkiawik",
+      description: "Wycieczka do Islandii",
+      start_date: new Date("2025-09-01"),
+      end_date: new Date("2025-09-15"),
+    },
+  });
```
AlicjaBonar: Your solution with create works correctly, but keep in mind it may fail if you run the seed multiple times and the records already exist. To avoid this, in the lecture it was shown to use the upsert command, which is usually preferred for seeding.

prisma/schema.prisma
```diff
@@ -0,0 +1,58 @@
...
+}
+
+model Expense {
+  expense_id  Int      @id @default(autoincrement())
+  amount      Float
+  category    Category
+  description String?
+  trip_id     Int
+  trip        Trip     @relation(fields: [trip_id], references: [trip_id])
+}
```
AlicjaBonar: It’s generally a good practice to include timestamps (createdAt and updatedAt) in data models. They are not always required, but they can be especially useful in models like Participant, Trip, and Expense. For other models, timestamps might also be helpful, though they are not strictly necessary. The ccorrect way to include them is shown in the lecture.

AlicjaBonar (approved):
Overall, great work! I noticed just two minor things that could be improved quickly to make it even better. Those issues are not critical for now, but it would help keep the data consistent and avoid problems later on

prisma/schema.prisma
feen1e: I understand, thank you!

PR #126, opened 2025-08-19
Jakub Wieśniak - Zadanie 2

MrChazar:
Hello, sorry for being late i had few things to do i hope you have still time to check my changes.


MrChazar:
I just noticed that I set the mass of fields to nullable, which doesn't make sense. I'll fix it tonight because it shouldn't be like that.

MrChazar:
I corrected those fields

MrChazar:
I will merge

Leadman5555:
Question: Can we merge by ourselves without a review?
Answer: No, we can not.
Question: Why did I not review your PR even after four days?
Answer: Because I didn't know about it - you did not request a review from me.

As I've mentioned before, if you do not request a review, I will not know about your PR, nor will I look for it among dozens of others. You have assigned me to the PR, but, as mentioned time and time again, that is not the same thing - I do not get an email when you assign me.

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+  trips     TripParticipant[] // M:N z Trip
+
+  @@map("Participant")
+}
+
+model Expanse {
+  id          Int      @id @default(autoincrement())
+  tripId      Int
+  amount      Int?
+  category    String?
```
Leadman5555, line 43: You did not use an enum type in your schema. It's important that you do, because enum types are very common, and yet they work differently than String or Int types - despite the fact that Prisma makes everything dumb and requires no real effort to make it work together, at the cost of somewhat decreased performance.

Replace the `category` column of "Expense" with an enum.

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+  participants TripParticipant[] // M:N z Participant
+
+  @@map("Trip")
+}
+
+model Participant {
+  id        Int              @id @default(autoincrement())
+  name      String?
+  surname   String?
+  role      Int?
```
Leadman5555, line 32: As for the `role` field, I'd replace it with an enum as well or remove it for now, as that field may change a bit after chapter 4 of the course.

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+
+  trips     TripParticipant[] // M:N z Trip
+
+  @@map("Participant")
+}
+
+model Expanse {
```
Leadman5555, line 39: You have a typo here. Best to change it early. "Expanse" is a different word than "Expense". "Expanse" means a vast area of something, "Expense" means an amount of something spent.

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+  id          Int      @id @default(autoincrement())
+  tripId      Int
+  amount      Int?
```
Leadman5555, line 42: Think about this field, should it be nullable? Does it make any sense for it to be optional? What is an expenditure with nothing spent?
Additionally, should it really be an integer?

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+  // relacje
+  expanses     Expanse[]         // 1 Trip -> wiele Expanses
+  participants TripParticipant[] // M:N z Participant
+
+  @@map("Trip")
+}
+
+model Participant {
+  id        Int              @id @default(autoincrement())
+  name      String?
```
Leadman5555, line 30: Those field should not be optional

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+
+model Trip {
+  id           Int               @id @default(autoincrement())
+  name         String?
```
Leadman5555, line 16: The name field at the very least should be required

prisma/schema.prisma
```diff
@@ -0,0 +1,61 @@
...
+  startDate    DateTime?
+  endDate      DateTime?
+  totalCost    Int?
```
Leadman5555, line 19: Should this really be an integer? Assuming it is a sum of all expenses, and expense can be a decimal, then the sum of decimals is also a decimal

prisma/seed.ts
```diff
@@ -0,0 +1,80 @@
...
+      totalCost: 3000,
+    },
+  });
+
+  await prisma.expanse.createMany({
+    data: [
+      {
+        tripId: trip.id,
+        amount: 1200,
+        category: 'Transport',
```
Leadman5555, line 36: Use enum types here - refer to my comment in the `schema` file

prisma/seed.ts
```diff
@@ -0,0 +1,80 @@
...
+  console.log('Sedowanie zakończone');
+}
+
+main()
+  .catch((e) => {
+    console.error(e);
+    process.exit(1);
+  })
+  .finally(async () => {
+    await prisma.$disconnect();
```
Leadman5555, line 79: This will not work how you would like it to. Calling an async method in an immediate termination context is a sure way to guarantee yourself undebuggable runtime errors. Refactor this termination sequence

src/database/database.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+export class DatabaseController {
+    constructor(private prisma: DatabaseService) {}
+
+    @Get('/Trip')
+    async getTrips(): Promise<Trip[]> {
+        return await this.prisma.trip.findMany();
+    }
+
+    @Post('/Trip')
+    async createTrip(@Body() tripData: Partial<Trip>): Promise<Trip> {
```
Leadman5555, line 16: This body type is not really correct. However, more of it will be explained in chapter 4 of the course, so you can leave it as is for now

Leadman5555 (commented):
Since this PR is already merged (although reverted), you need to create a new one. See my comments, implement fixes, and open a new one with the fixed version.

## Remember to request review or I will not know about it

MrChazar:
Sorry for merging too soon i must find where i can inform you about review i will repair my code as soon as possible.

Leadman5555:
image

You request review in the Reviewers Tab (top right corner of the PR view). "Assignees" means people assigned to **do** the PR - that's you. "Reviewers" are people assigned to do the review - that's me.
If you add me as an "Reviewer", I receive an email about it.

PR #128, opened 2025-08-20
Jakub Morawiec - zadanie 2

jmorawiec:
Wybacz za opóźnienie - byłem na wyjeździe, a fakt, że prowadzę WW z promo sprawiło, że niestety priorytetem było prowadzenie mojego kursu. W każdym razie podsyłam pliki do sprawdzenia, w większości nie było problemów z wykonaniem zadanka.


dowód że działa.png
Leadman5555, line 1: In the future, post the photo in the PR description or in a PR comment. Don't upload it to the codebase.

prisma/schema.prisma
```diff
@@ -0,0 +1,76 @@
...
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+// enums
+enum TripStatus {
+  planned
```
Leadman5555: Enum naming. By convention, enums are nearly always written in full caps lock snake case (that is: `MY_ENUM`). It's mostly due to "tradition" - in C all constants were defined in this way. Thus, whenever someone sees a variable like "STATUS_COMPLETED", they are likely to think that it's an constant/enum. It's not a big mistake to name them differently, but the convention is so common that it just feels unusual.

prisma/seed.ts
```diff
@@ -0,0 +1,66 @@
...
+import { PrismaClient } from "../generated/prisma";
+
+const prisma = new PrismaClient();
+
+async function main() {
+  const trip = await prisma.trip.create({
+    data: {
+      name: "Kajaki i namioty",
+      description: "Wyjazd na spływ kajakowy z noclegami w namiotach.",
+      status: "planned",
```
Leadman5555: I don't like the magic values here and with all other enum strings. My suggestion is to create a file with the enum (mirroring the one in the prisma schema) and to use the constants from the file: `status: TripStatus.PLANNED`. The current approach with the hand-typed and hard coded strings is just too error prone

prisma/seed.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+  });
+}
+
+main()
+  .catch((e) => {
+    console.error("didnt work, reason:", e);
+    process.exit(1);
+  })
+  .finally(async () => {
```
Leadman5555, line 69: Nice touch with the disconnect in `finally` 👍

zadanie1.png
Leadman5555, line 1: Same as with the first photo - in the future, post this in the PR description or comment.

Leadman5555 (changes requested):
The code, apart from a few minor suggestions, looks good. **But**... there is one critical problem - the relationships between Trip, Participant and Expense.

Why? Well, we have a relationship loop of sorts. Please look at the example below:

----------------------
Participant:

1. 	paricipant_id = 1
	trip_id = 1

Trip:

1.	trip_id = 1
2.	trip_id = 2

Expense:

1. 	trip_id = 2
	participant_id = 1
	
----------------------

So, we have a participant that is on trip 1 and paid for trip 2. Of course, you can write code in a such a way that this will not occur. But you shouldn't have to worry about it. Such behaviour should be forbidden by the database itself. As such we have two solutions:
1. Add constraints (a quick fix, but not what we will do here)
2. Change the relationships (the proper approach since we are not too deep yet)

TLDR: change the relationships so that there is no cycle.

Leadman5555:
> Wybacz za opóźnienie - byłem na wyjeździe, a fakt, że prowadzę WW z promo sprawiło, że niestety priorytetem było prowadzenie mojego kursu. W każdym razie podsyłam pliki do sprawdzenia, w większości nie było problemów z wykonaniem zadanka.

Absolutely no need to apologise, especially since you have a course of your own to lead. In fact, you are the second person sending me a part 2 PR, so congrats on that 👍

prisma/seed.ts
Leadman5555: Or, alternatively, use the prisma enum support. Either way, do not hardcode the values

zadanie1.png
Leadman5555, line 1: This is outdated as you have updated your schema. Either update the photo as well or remove it - we cannot leave the old version here as it is misleading.

prisma/schema.prisma
```diff
@@ -0,0 +1,69 @@
...
+  @@index([tripId])
+}
+
+model Expense {
+  id        Int              @id @default(autoincrement())
+  payerId   Int
+  amount    Decimal          @db.Decimal(12, 2)
+  category  ExpenseCategory
+  currency  String           @db.VarChar(3)
+  paid_at   DateTime?
```
Leadman5555, line 66: We are mixing the naming case. Some columns are in camelCase (`payerId`, `startDate`), some are in snake_case (`payed_at`, `joined_at`). Is it an code bug? No. But we should stick to one convention. Either use snake_case in all fields or camelCase - your choice, but choose one and use it.

Leadman5555 (approved):
Overall, you fixed the errors. I left two comments on minor things, but will nevertheless approve this PR. Once you fix the two things, you can merge it by yourself - I trust you will not break anything when changing the naming cases 👍

PR #131, opened 2025-08-20
Jakub Ptak zadanie 2

ptakj:
Sorry for being late, I`ve encountered some PrismaClient problems that took much more time that it suposed to.


Leadman5555 (changes requested):
Three things:

1. Stop force pushing.
2. I've already told you before, to name your commits appropriately - if you continue to name all of them the same way, I will not accept your PRs. Please refer to the provided commit naming guide (linked in the course readme)
3. The task is not done according to the instructions - you are missing several requirements. Please reread the instructions and send a fixed version.

Leadman5555 (changes requested):
Three things:

1. Stop force pushing.
2. I've already told you before, to name your commits appropriately - if you continue to name all of them the same way, I will not accept your PRs. Please refer to the provided commit naming guide (linked in the course readme)
3. The task is not done according to the instructions - you are missing several requirements. Please reread the instructions and send a fixed version.

baza.png
Leadman5555, line 1: In the future, send images like this in the PR description. No need to put in into the code.

prisma/schema.prisma
```diff
@@ -0,0 +1,47 @@
...
+  trip_id Int
+  gender  Gender
+  trip  Trip  @relation(fields: [trip_id], references: [id])
+}
+
+model Trip {
+  id    Int     @default(autoincrement()) @id
+  date   DateTime
+  location  String
+  duration Int
```
Leadman5555: The `duration` field is a bit ambiguous. If I understand it correctly, it should mean something along the lines of "how many days does a trip last". But still, it's not intuitive. Replace it with an `endDate` field. Also, rename `date` to `startDate`

prisma/schema.prisma
```diff
@@ -0,0 +1,47 @@
...
+  participant Participant[]  
+  expense     Expense[] 
+}
+
+model Expense {
+  id    Int     @default(autoincrement()) @id
+  daily_price   Float
+  trip_id  Int
+  discount Boolean?
```
Leadman5555: I'm not sure what does the `discount` field represent. Could you explain it in the reply to this comment?

prisma/seed.ts
```diff
@@ -0,0 +1,44 @@
...
+      duration: 10,
+    },
+  });
+
+  const part = await prisma.participant.create({
+    data: {
+      name: "Jan",
+      surname: "Kowalski",
+      age: 25,
+      gender: "M",
```
Leadman5555: Use the `Gender` enum here. That's their point - to remove hardcoded constant values.

prisma/seed.ts
```diff
@@ -0,0 +1,44 @@
...
+  });
+}
+
+main()
+  .then(async () => {
+    await prisma.$disconnect();
+  })
+  .catch(async (e) => {
+    console.error(e);
+    await prisma.$disconnect();
```
Leadman5555, line 42: This, in stark contrast to the implementations of other people, actually is correct 👍 . However, there is a cleaner way to do the same thing. See the `process.exitCode` property - you don't need to manually call the `exit()` method then and can move the `disconnect()` call to a "finally" block

Leadman5555 (changes requested):
Several minor issues with the schema columns. There is one question in the comments, be sure to answer it 👍

Leadman5555:
image

There is not need to do all that 🤣 .  As I've mentioned time and time again, before chapter with tests, you can simple outright delete the test files.

prisma/schema.prisma
ptakj: Often company provides constant percentage discount, so only thing that we need to storage in DB is presence or absence of that discount. On top of that it is a way to add Boolean field to DB and use required optional field.

Leadman5555: Alright, I understand what you mean. Although, `discount = true` means there is one, `discount = false` means there is not, so what does `discount = null` mean? I'd assume null and false have the same meaning, but that's not really intuitive. But since we do need an optional field or two, I'd suggest the following changes:
- make the `discount` required
- make Participant `age` or `gender` optional

prisma/seed.ts
```diff
@@ -0,0 +1,47 @@
...
+      discount: true,
+    },
+  });
+}
+
+main()
+  .then(async () => {
+    await prisma.$disconnect();
+  })
+  .catch(async (e) => {
```
Leadman5555, line 37: This is not really what I meant by my previous comment - the new version looks ugly 😢.  There is no need to guard the `disconnect` too.
I meant something along the lines of:
```ts
main()
  .catch(async (e) => {
    console.error(e);
    process.exitCode = 1;
  }).finally(async () => {
    await prisma.$disconnect();
});
```
It's a cleaner solution 👍

prisma/schema.prisma
```diff
@@ -0,0 +1,47 @@
...
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+model Participant {
+  id    Int     @default(autoincrement()) @id
+  name String  
+  surname  String
+  age Int
+  trip_id Int
```
Leadman5555: We are mixing the naming case. Some columns are in camelCase (`endDate`, `startDate`), some are in snake_case (`daily_price`, `trid_id`). Is it an code bug? No. But we should stick to one convention. Either use snake_case in all fields or camelCase - your choice, but choose one and use it.

src/database/database.controller.ts
```diff
@@ -0,0 +1,24 @@
+import { Controller, Get } from "@nestjs/common";
+import { Expense, Participant, Trip } from "@prisma/client";
+
+import { DatabaseService } from "./database.service";
+
+@Controller("database")
+export class DatabaseController {
+  constructor(private prisma: DatabaseService) {}
+
+  @Get("/Trip")
```
Leadman5555, line 16: Change the URI in all endpoints to lowercase. When dealing with URI, the usual convention is for them to in kebab-case

src/database/database.controller.ts
```diff
@@ -0,0 +1,24 @@
+import { Controller, Get } from "@nestjs/common";
+import { Expense, Participant, Trip } from "@prisma/client";
+
+import { DatabaseService } from "./database.service";
+
+@Controller("database")
+export class DatabaseController {
+  constructor(private prisma: DatabaseService) {}
```
Leadman5555, line 8: Let's not mix different app layers (web and persistence). It has been, or should have been, mentioned in the 3rd chapter of the course (last week one). Create a TripService and move the database calls there.

src/database/database.controller.ts
```diff
@@ -0,0 +1,24 @@
+import { Controller, Get } from "@nestjs/common";
+import { Expense, Participant, Trip } from "@prisma/client";
+
+import { DatabaseService } from "./database.service";
+
+@Controller("database")
```
Leadman5555, line 6: There should be a separate controller per resource

Leadman5555 (changes requested):
Some minor fixes to do. As for the CRUD part, I know that it is mainly done in the 3rd task, not the 2nd one, but since you started, you might as well do it properly - it will mean less work in the next task for you 👍

Leadman5555 (approved):

PR #136, opened 2025-08-23
Dominik dabrowski 2

DominikDab341:


Leadman5555:
Why are we trying to merge into the `main` branch again? I'll change it for you this time, but please remember to choose the correct target yourself

prisma/schema.prisma
```diff
@@ -0,0 +1,62 @@
...
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+enum Role {
+  member
+  organizer
+}
+
+enum ExpenseType {
```
Leadman5555, line 21: Enum naming. By convention, enums are nearly always written in full caps lock snake case (that is: MY_ENUM). It's mostly due to "tradition" - in C all constants were defined in this way. Thus, whenever someone sees a variable like "STATUS_COMPLETED", they are likely to think that it's an constant/enum. It's not a big mistake to name them differently, but the convention is so common that it just feels unusual. Just change them all to the correct case

prisma/schema.prisma
```diff
@@ -0,0 +1,62 @@
...
+generator client {
+  provider = "prisma-client-js"
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+enum Role {
```
Leadman5555, line 16: Same thing here. Change the case

prisma/schema.prisma
```diff
@@ -0,0 +1,62 @@
...
+
+
+}
+
+model Expense {
+  expense_id   Int         @id @default(autoincrement())
+  trip_id      Int
+  trip         Trip        @relation(fields: [trip_id], references: [trip_id])
+  expense_type ExpenseType
+  expense_date DateTime    @db.Date
```
Leadman5555: With money-related matters, it's usually better to be more detailed than not. Using the `@db.Date` you narrow the type down to just `DATE` (without the time part). While might be a suitable approach with the trip dates, when dealing with expenses, we should be more detailed. Simply remove the restriction and the column should map to `TIMESTAMP` instead of `DATE`

Leadman5555 (changes requested):
Very minor changes to make. Incidentally, you were the only person so far to using the column restrictions in the schema - and it's a good thing that you did 👍 .
When you fix the enum case and `expense_date` type, you can merge the PR - I trust that your fixes will not need a second review

Leadman5555:
image

As for the failing checks, you can actually see what failed in detail. When pushing the fixes, run prettier locally before committing. You can remove the failing test for now, tests are not required until later on in the course

Leadman5555 (changes requested):
I think you misunderstood me. I wrote about changing the case of the enum values. Which is , for example "food" -> "FOOD". Not the enum name itself. Revert the name to what they were (Role, ExpenseType), and change the values. After all, they are the constants.

DominikDab341:
I will change it, sorry about that. I have a question regarding the tests: are they failing because GitHub doesn’t see the PostgreSQL DATABASE_URL? And how can I fix this properly?

Leadman5555:
Well, the `DATABASE_URL` is read from the environment variables. There is no .env file to inject from and there are no variables specified in the workflow either. Even it there was a `DATABASE_URL` variable present, is there any database to connect to? Workflow run in an isolated environment and since there is no database service added there, there is no database to possibly connect to.
The tests fail not because of your code but because the CI is not supporting database connection **yet**. It will be changed when we get to the testing chapter of the course, so don't worry about it for now.

PR #139, opened 2025-08-24
Hubert drewlo - task 3

Drewienkopl:
I have made fixes according to what was pointed out and everything should now be good.

swaggerWorks
swaggerWorks2
swaggerWorks3
swaggerWorks4
swaggerWorks5
swaggerWorks6
swaggerWorks7


Leadman5555:
You need to finish part 2 first before starting part 3.

Leadman5555:
> You need to finish part 2 first before starting part 3.

That is: #102

src/expense/dto/create-expense.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+  @IsPositive()
+  amount!: number;
+
+  @ApiProperty({ enum: ExpenseCategory, enumName: "ExpenseCategory" })
+  @IsEnum(ExpenseCategory)
+  category!: ExpenseCategory;
+
+  @ApiProperty({ required: false, example: "Train tickets" })
+  @IsOptional()
+  @ApiPropertyOptional()
```
Leadman5555: Stacking both `@ApiPropertyOptional` and `@ApiProperty` makes the autogenerated definition behave in a weird way. Choose only one of them. Also, your linter should have warned you about it

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Post()
+  @ApiOperation({ summary: "Create a new expense" })
+  @ApiCreatedResponse({ description: "Created expense" })
+  async create(@Body() dto: CreateExpenseDto) {
+    return this.service.create(dto);
+  }
+
+  @Patch(":id")
```
Leadman5555, line 55: You are missing the operation tags for Swagger. Add them to **every** controller endpoint. Include the following two fields:
- summary
- return value schema

src/expense/expense.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (item === null) {
+      throw new NotFoundException("Expense not found");
+    }
+    return item;
+  }
+
+  async create(dto: CreateExpenseDto) {
+    return this.prisma.expense.create({
+      data: {
+        participantId: dto.participantId,
```
Leadman5555, line 40: What if the participant of the given Id does not exist? We need to check that first

src/expense/expense.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  async update(id: number, dto: UpdateExpenseDto) {
+    await this.ensureExists(id);
+    return this.prisma.expense.update({
+      where: { id },
+      data: {
+        amount: dto.amount,
+        category: dto.category,
+        note: dto.note,
+        participantId: dto.participantId,
```
Leadman5555, line 74: Same thing here

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  @ApiOkResponse({ description: "Updated expense" })
```
Leadman5555, line 57: Is 200 the only possible return code you can get here? After changing the service methods as requested, think what other codes should go here. Of course, omit the 500

src/participant/dto/create-participant.dto.ts
```diff
@@ -0,0 +1,37 @@
+import { ParticipantRole } from "@prisma/client";
+import { IsEmail, IsEnum, IsInt, IsOptional, IsString } from "class-validator";
+
+import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
+
+export class CreateParticipantDto {
```
Leadman5555, line 6: Same here. Fix the ApiProperty decorators

src/participant/dto/create-participant.dto.ts
```diff
@@ -0,0 +1,37 @@
...
+  @ApiProperty({ example: "Marek" })
+  @IsString()
+  name!: string;
```
Leadman5555: Quick question, what is the exclamation mark (!) doing here? First of all, it might be treated like a normal character here. And even if it were not, it should not be here.
Overall, the only place you should use the non-null operator is when you are 100% sure that the value is not nullish, you have checked it, but the linter still does not understand it - such situations happen. And even then, you should think twice whether the linter really is wrong.

src/participant/participant.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  ApiTags,
+} from "@nestjs/swagger";
+
+import { CreateParticipantDto } from "./dto/create-participant.dto";
+import { UpdateParticipantDto } from "./dto/update-participant.dto";
+import { ParticipantService } from "./participant.service";
+
+@ApiTags("participants")
+@Controller("participants")
+export class ParticipantController {
```
Leadman5555, line 26: Same things to fix as in the previous controller

src/trip/trip.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+  @ApiOkResponse({ description: "List of trips" })
+  async getAllTrips() {
+    return this.tripService.getAllTrips();
+  }
+
+  @Post()
+  @ApiOperation({ summary: "Create a new trip" })
+  @ApiOkResponse({ description: "The created trip" })
+  async createTrip(
+    @Body() body: { name: string; startDate: string; budget?: number },
```
Leadman5555: Replace this with a proper DTO.

Leadman5555 (changes requested):
A few things to fix here. Details in comments

Leadman5555 (approved):

PR #157, opened 2025-08-27
feat/ Dominik kaczmarek lista 3

feen1e:
Sorry for being late - my computer broke on friday and I managed to get it fixed today.

Added CRUD and DTO for all tables as well as Swagger decorators.
image
image

### DIFF
--- a/src/trip-participants/trip-participants.service.ts
+++ b/src/trip-participants/trip-participants.service.ts
@@ -1,0 +1,160 @@
+import {
+  ConflictException,
+  Injectable,
+  NotFoundException,
+} from "@nestjs/common";
+
+import { DatabaseService } from "../database/database.service";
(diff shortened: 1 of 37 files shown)


src/trip-participants/trip-participants.service.ts
```diff
@@ -0,0 +1,152 @@
...
+          trip_id: tripId,
+          participant_id: participantId,
+        },
+      });
+
+    if (tripParticipant === null) {
+      throw new NotFoundException("Trip participant not found");
+    }
+
+    await this.databaseService.tripParticipants.deleteMany({
```
AlicjaBonar: Here you could use delete instead of delete many. You want to delete just one object anyway so it would be safer

AlicjaBonar (approved):
Very good work! I just added one small suggestion. No worries about the delay

PR #163, opened 2025-08-30
Hubert drewlo - task 4

Drewienkopl:
I made it so all GET, register and login are public, POST and PATCH aren't, and trips POST and PATCH are only for admins/cordinators.

This is result of this task: 

You can GET without login:
1_expenceGetWorks

You can't POST unauthorised:
2_expencePostWorksNot

Registering an account works:
_register1Works

You can't create second account with the same email, it gives conflict error:
_register1NotWorksConflict

You can login:
_login1Works

Wrong password = you can't login and give "Incorrect password" message with unauthorized error:
_login1WorksNOT

Inputting not existing user email gives "User not found" message:
_login2WorksNOT

After logging in using token, now it's possible to POST:
POST_WorksWhenLoggedIn

Putting in wrong token gives "Invalid token message":
InvalidToken

Changing your own data works:
ChangingYourNameWorks

You can't change another account's data:
OnlyYourOwn

Can't change role:
image

Validator works:
ValidatorWorks

Admin/Cordinator can create new trips:
adminCan

Users can't create new trips:
UsersCant


Leadman5555:
> I have been forgetting to do merges of previous tasks, but I have been making fixes to them as per comments, I will clean up the previous tasks merges and everything after weekend as I have exam on monday and I really should start studying for it

@Drewienkopl, First of all, I haven't even graded your 3rd task and your 2nd PR is still open. Your first branch isn't even merged. This is not how it works. There is no point in doing the 4th chapter if your first one still is not merged

What you should do is:
1. merge the first branch 
2. make fixes to the **2nd** (that is #102 ), according to my comments. Do it in the 2nd PR. Not in this one, or any other one. Request review again after fixing.
3. Wait for me to approve your fixes
4. If my review is negative, go back to step 2. Otherwise, if I approve, merge the 2nd PR into your base branch and go to the next step
5. Wait for my review on your third PR and continue from there. 

I am closing this PR. The course is divided into chapters (stages) and should be done in stages.

PR #165, opened 2025-08-31
Zadanie 3 - Jakub Morawiec

jmorawiec:
Trochę się z tym męczyłem, ale się udało. CRUDy przetestowane z moją bazą danych, działają.


jmorawiec:
All checks are done, @Leadman5555 please check

src/expenses/expenses.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Get()
+  @ApiOperation({ summary: "Pobierz wszystkie wydatki" })
+  @ApiOkResponse({ description: "Zwraca listę wydatków" })
+  async findAll() {
+    return this.expensesService.findAll();
+  }
+
+  @Get(":id")
+  @ApiOperation({ summary: "Pobierz wydatek" })
+  @ApiOkResponse({ description: "Zwraca wydatek" })
```
Leadman5555: Here it would be nice to include other possible responses, for example:
```ts
@ApiResponse({
    status: 404,
    description: "Expense not found",
  })
```

src/expenses/expenses.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  async findOne(@Param("id", ParseIntPipe) id: number) {
+    return this.expensesService.findOne(id);
+  }
+
+  @Patch(":id")
+  @ApiOperation({ summary: "Aktualizuj wydatek" })
+  @ApiOkResponse({ description: "Zaktualizowano wydatek" })
```
Leadman5555: Add types to response Swagger definitions where applicable. That is:
```ts
@ApiResponse({
    type: [YourReturnTypeHere],
  })
```

src/participants/participants.controller.ts
Leadman5555, line 1: Same here, add other responses to endpoints. And types.

Leadman5555 (changes requested):
Some Swagger tags are missing, but overall looks okay. Writing CRUD is mainly boilerplate so as long as you are happy with how the endpoints work currently, then there is no problem. 
You needn't have to have added validators on the fields - it's required in the 4th chapter. But since you already did, you will have less work in the future 👍 
So add the required decorators and I think we will be good to go.

src/expenses/entities/expense.entity.ts
Leadman5555, line 1: Let's remove this file. Since we are using Prisma as our ORM, it generates the entities for us (you can import them from prisma) . Do note, that most ORMs **do not** do that for many various reason. But, since this is a learning course, Prisma was chosen, entity generation, being one of the reasons for it.
Even if you define the entity here, Prisma will still use its own auto-generated version. So it's not only dead code, but also can lead to errors if your definition differs from the generated one and you structure your code to fit your own definition

Leadman5555 (approved):
I'll accept it. It's just boilerplate, so no need to spend a lot of time on it. Remove me `expense.entity.ts` file. After that, you can merge the PR 👍

src/expenses/entities/expense.entity.ts
jmorawiec: I changed the code and faced the error: ''Expense' only refers to a type, but is being used as a value here.' after importing it from Prisma.

Leadman5555: Well, the error suggests that you were using it as a value. For example, it could have been: `const whatever = Expense`, while it should be `const whatever: Expense = ...`

jmorawiec: Weird thing is that it looks like this in my code:
`  @ApiOkResponse({
    description: "Zwraca listę wydatków",
    type: [Expense],
  })`
  
  `
  { description: "Zwraca wydatek", type: Expense }
  `
  
  but the error still occurs

PR #168, opened 2025-08-31
Jakub Ptak  - zadanie 3

ptakj:


src/expense/entities/expense.entity.ts
Leadman5555, line 1: There is no need to define entities when using Prisma. Remove this file.

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  Post,
+} from "@nestjs/common";
+import { ApiOperation, ApiResponse } from "@nestjs/swagger";
+
+import { CreateExpenseDto } from "./dto/create-expense.dto";
+import { UpdateExpenseDto } from "./dto/update-expense.dto";
+import { ExpenseService } from "./expense.service";
+
+@Controller("expense")
+export class ExpenseController {
```
Leadman5555, line 26: Add Swagger tags for the controller itself

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  constructor(private readonly expenseService: ExpenseService) {}
+
+  @Post()
+  @HttpCode(HttpStatus.CREATED)
+  @ApiOperation({
+    summary: "Created a new expense",
+    description: "Added a expense",
```
Leadman5555: Description is a full version of summary. It is optional, but if you do include it, it should be what it is supposed to be.

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Get()
+  @HttpCode(HttpStatus.CREATED)
+  @ApiOperation({
+    summary: "Returning all expenses",
+    description: "Returned all expenses!",
+  })
+  @ApiResponse({
+    status: 201,
```
Leadman5555: 201 CREATED is not the status code we should have here

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+    description: "Expenses returned",
+  })
+  async findAll() {
+    return this.expenseService.findAll();
+  }
+
+  @Get(":id")
+  @HttpCode(HttpStatus.CREATED)
```
Leadman5555: Same thing here

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  @ApiResponse({
+    status: 201,
+    description: "Expense found!",
+  })
+  async findOne(@Param("id") id: string) {
+    return this.expenseService.findOne(+id);
+  }
+
+  @Patch(":id")
+  @HttpCode(HttpStatus.CREATED)
```
Leadman5555: And here

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  })
+  async update(
+    @Param("id") id: string,
+    @Body() updateExpenseDto: UpdateExpenseDto,
+  ) {
+    return this.expenseService.update(+id, updateExpenseDto);
+  }
+
+  @Delete(":id")
+  @HttpCode(HttpStatus.CREATED)
```
Leadman5555: And with all other endpoints. I will not comment on it anymore

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+    return this.expenseService.findAll();
+  }
+
+  @Get(":id")
+  @HttpCode(HttpStatus.CREATED)
+  @ApiOperation({
+    summary: "Found an expense with given id",
+    description: "Found an expense with given id",
+  })
+  @ApiResponse({
```
Leadman5555, line 71: What if the `find` fails? There is no response defined for that, is there?

src/expense/expense.service.ts
```diff
@@ -0,0 +1,43 @@
...
+      },
+    });
+  }
+
+  async findAll() {
+    return this.database.expense.findMany();
+  }
+
+  async findOne(id: number) {
+    return this.database.expense.findUnique({ where: { id } });
```
Leadman5555: If the query finds nothing, it will return `null`. In the current implementation that behavoiur is treated as a correct one. This is wrong. Failure to find should result in a failure status code, not a success one

src/expense/expense.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  async update(id: number, updateExpenseDto: UpdateExpenseDto) {
+    return this.database.expense.update({
```
Leadman5555, line 56: Same thing here and in all other methods that have the same structure. What if there is no expense of the given `id`?

src/participant/dto/create-participant.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+  surname: string;
+
+  @ApiProperty()
+  age: number;
+
+  @ApiProperty()
+  tripId: number;
+
+  @ApiPropertyOptional()
+  gender?: Gender;
```
Leadman5555, line 22: If it's an enum, you should include that in the Swagger definition.

src/participant/entities/participant.entity.ts
Leadman5555, line 1: There is no need to define entities when using Prisma. Remove this file.

src/participant/participant.controller.ts
Leadman5555, line 1: The exact same problems as in the ExpenseController. I will not write explicit comments here, but it needs to be fixed as well

src/participant/participant.service.ts
Leadman5555, line 1: The exact same problems as in ExpenseService. Fix them all as well

src/trip/entities/trip.entity.ts
Leadman5555, line 1: There is no need to define entities when using Prisma. Remove this file.

src/trip/trip.controller.ts
Leadman5555, line 1: The exact same problems as in the ExpenseController. I will not write explicit comments here, but it needs to be fixed as well

src/trip/trip.service.ts
Leadman5555, line 1: The exact same problems as in ExpenseService. Fix them all as well

Leadman5555 (changes requested):
Well, a lot of things to fix. Most of them repeated mistakes in the service and controller files.

src/expense/entities/expense.entity.ts
ptakj: Removed

src/expense/expense.controller.ts
ptakj: @ApiTags("Expense controller") added

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,139 @@
...
+    status: 200,
+    description: "Expense found",
+    type: ResponseExpenseDto,
+  })
+  @ApiResponse({
+    status: 404,
+    description: "Expense not found",
+  })
+  @ApiResponse({
```
Leadman5555: Including the 500 code in documentation is kind of pointless. If you get 500, it means that whatever happened on the server was not accounted for. That nobody thought that it could happen. As long as you know something can happen, it should not be 500.

src/expense/expense.service.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+    if (expense == null) {
+      throw new NotFoundException("Expense not found");
+    } else {
+      return expense;
+    }
+  }
+
+  async update(id: number, updateExpenseDto: UpdateExpenseDto) {
+    const expense: unknown = this.database.expense.findUnique({
```
Leadman5555: Extract this code (find one and throw if none found) to a separate method. You can call it `findOneOrFail`, `findOrThrow` or whatever you feel like. That way the same code snippet will not repeat

src/expense/expense.service.ts
```diff
@@ -0,0 +1,64 @@
...
+      where: { id },
+    });
+    if (expense == null) {
+      throw new NotFoundException("Expense not found");
+    } else {
+      return this.database.expense.update({
+        where: { id },
+        data: {
+          dailyPrice: updateExpenseDto.dailyPrice,
+          tripId: updateExpenseDto.tripId,
```
Leadman5555: What is the trip of the given `tripId` does not exist? We should check that first.

src/participant/participant.controller.ts
```diff
@@ -0,0 +1,133 @@
...
+  @ApiResponse({
+    status: 201,
+    description: "Created a new participant",
+    type: CreateParticipantDto,
+  })
+  @ApiResponse({
+    status: 400,
+    description: "Invalid participant data",
+  })
+  @ApiResponse({
```
Leadman5555: Same here with the 500 codes

src/participant/participant.service.ts
```diff
@@ -0,0 +1,68 @@
...
+export class ParticipantService {
+  constructor(private database: DatabaseService) {}
+
+  async create(createParticipantDto: CreateParticipantDto) {
+    return this.database.participant.create({
+      data: {
+        name: createParticipantDto.name,
+        surname: createParticipantDto.surname,
+        age: createParticipantDto.age,
+        tripId: createParticipantDto.tripId,
```
Leadman5555, line 36: Check if the trip exists first

src/participant/participant.service.ts
```diff
@@ -0,0 +1,68 @@
...
+          age: updateParticipantDto.age,
+          tripId: updateParticipantDto.tripId,
+          gender: updateParticipantDto.gender,
+        },
+      });
+    }
+  }
+
+  async remove(id: number) {
+    const participant: unknown = this.database.participant.findUnique({
```
Leadman5555: Same as in the previous service. Extract the find one or throw snippet

src/expense/expense.controller.ts
Leadman5555: As in the comment in the previous PR review. We should not return a CREATED code on a DELETED resource

Leadman5555 (changes requested):
A few things to fix. Also, you missed a comment or two from the previous review, so check up on it.

src/expense/expense.controller.ts
ptakj: Ok, code 500 will be removed

src/expense/expense.controller.ts
ptakj: Oh sorry it seems like I have not saved file, my bad.

Leadman5555 (approved):

PR #175, opened 2025-09-02
Dominik dabrowski 3

DominikDab341:


src/expense/dto/create-expense.dto.ts
```diff
@@ -0,0 +1,52 @@
...
+  })
+  @IsEnum(ExpenseType)
+  expense_type: ExpenseType;
+
+  @ApiProperty({
+    description: "Date when the expense occurred (YYYY-MM-DD)",
+    example: "2025-09-01",
+  })
+  @IsDateString()
+  expense_date: string;
```
Leadman5555: This should be a date, not a string. You can use a @Type decorator to convert it

src/expense/dto/expense-response.dto.ts
```diff
@@ -0,0 +1,34 @@
...
+    description: "Unique identifier of the expense",
+    example: 1,
+  })
+  expense_id: number;
+
+  @ApiProperty({
+    description: "Type of the expense",
+    example: "FOOD",
+  })
+  expense_type: string;
```
Leadman5555, line 14: According to your schema, this is an enum, not a string. You should reflect it in types and in Swagger

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+  ) {
+    return this.expenseService.update(+id, updateExpenseDto);
+  }
+
+  @Delete(":id")
+  @ApiOperation({ summary: "Delete expense by ID" })
+  @ApiResponse({ status: 200, description: "Expense deleted successfully" })
+  @ApiResponse({ status: 404, description: "Expense not found" })
+  async remove(@Param("id") id: string) {
+    return this.expenseService.remove(+id);
```
Leadman5555: Returning a message on delete is a bit pointless. Either return nothing with 204 or return the deleted entity with 200

src/expense/expense.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+@ApiTags("expenses")
+@Controller("expense")
+export class ExpenseController {
+  constructor(private readonly expenseService: ExpenseService) {}
+
+  @Post()
+  @ApiOperation({ summary: "Create a new expense" })
+  @ApiResponse({ status: 201, description: "Expense created successfully" })
+  async create(@Body() createExpenseDto: CreateExpenseDto) {
```
Leadman5555, line 33: Are you sure 201 CREATED is the only possible response here? Look at this endpoint and at all other endpoints to see if you have included all of the possible responses. Of course, 500 is skipped.

src/expense/expense.service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+import { CreateExpenseDto } from "./dto/create-expense.dto";
+import { UpdateExpenseDto } from "./dto/update-expense.dto";
+
+@Injectable()
+export class ExpenseService {
+  constructor(private readonly database: DatabaseService) {}
+
+  async create(createExpenseDto: CreateExpenseDto) {
+    const trip = await this.database.trip.findUnique({
```
Leadman5555: Extract the findUnique calls with null checks to a single method

src/expense/expense.service.ts
```diff
@@ -0,0 +1,129 @@
...
+      where: { trip_id: createExpenseDto.trip_id },
+    });
+
+    if (trip == null) {
+      throw new NotFoundException(
+        `Trip with ID ${String(createExpenseDto.trip_id)} not found`,
```
Leadman5555: Don't use `String(numer)`. Use `number.toString()` instead.

src/expense/expense.service.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (trip == null) {
+        throw new NotFoundException(
+          `Trip with ID ${String(updateExpenseDto.trip_id)} not found`,
+        );
+      }
+    }
+
+    // jawny null/empty-check dla expense_date (bez truthy-checka)
+    const nextExpenseDate =
+      updateExpenseDto.expense_date == null ||
```
Leadman5555: Assuming you correct the DTO, the date conversion will not be needed.

src/participant/participant.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+    @Body() updateParticipantDto: UpdateParticipantDto,
+  ) {
+    return this.participantService.update(+id, updateParticipantDto);
+  }
+
+  @Delete(":id")
+  @ApiOperation({ summary: "Delete participant by ID" })
+  @ApiResponse({ status: 200, description: "Participant deleted successfully" })
+  @ApiResponse({ status: 404, description: "Participant not found" })
+  async remove(@Param("id") id: string) {
```
Leadman5555: Same thing here, with message on delete.

src/participant/participant.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      throw new NotFoundException(
+        `Participant with ID ${String(id)} not found`,
+      );
+    }
+
+    return participant;
+  }
+
+  async update(id: number, updateDto: UpdateParticipantDto) {
+    const participant: Participant | null =
```
Leadman5555: Extract the findUnique calls with null checks to a single method

src/trip/dto/create-trip.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+  })
+  @IsString()
+  destination: string;
+
+  @ApiProperty({
+    description: "Start date of the trip (YYYY-MM-DD)",
+    example: "2025-09-01",
+  })
+  @IsDateString()
+  start_date: string;
```
Leadman5555: Here as well, this should be a properly transformed date

src/trip/dto/create-trip.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  @ApiPropertyOptional({
+    description: "End date of the trip (YYYY-MM-DD)",
+    example: "2025-09-03",
+  })
+  @IsOptional()
+  @IsDateString()
+  end_date?: string | null;
```
Leadman5555: string | null | undefined ? Either use null or use undefined. Stacking both in one place, at least in this situation, leads to bugs. null means no date, undefined... well, you didn't implement that. Although, it should fallback to no changes to the date on default, which can lead to database errors on create.

src/trip/dto/create-trip.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  @ApiPropertyOptional({
+    description: "Planned budget for the trip",
+    example: 1200,
+  })
+  @IsOptional()
+  @IsNumber()
+  budget?: number | null;
```
Leadman5555: Same thing here with null | undefined

src/trip/trip.controller.ts
```diff
@@ -0,0 +1,112 @@
...
+  async update(@Param("id") id: string, @Body() updateTripDto: UpdateTripDto) {
+    return this.tripService.update(+id, updateTripDto);
+  }
+
+  @Delete(":id")
+  @ApiOperation({
+    summary: "Delete a trip",
+    description: "Remove a trip and all its associated data from the system",
+  })
+  @ApiResponse({
```
Leadman5555, line 101: Same thing with message on delete.

src/trip/trip.service.ts
```diff
@@ -0,0 +1,93 @@
...
+  constructor(private database: DatabaseService) {}
+
+  async create(createTripDto: CreateTripDto) {
+    return this.database.trip.create({
+      data: {
+        name: createTripDto.name,
+        destination: createTripDto.destination,
+        start_date: new Date(createTripDto.start_date),
+        end_date:
+          createTripDto.end_date == null
```
Leadman5555: And what if it's undefined? Change the DTO definition.

src/trip/trip.service.ts
```diff
@@ -0,0 +1,93 @@
...
+        destination: updateTripDto.destination,
+        start_date: updateTripDto.start_date,
+        end_date: updateTripDto.end_date,
+        budget: updateTripDto.budget,
+      },
+    });
+  }
+
+  async remove(id: number) {
+    const trip = await this.database.trip.findUnique({
```
Leadman5555: Extract the `findUnique` calls here as well

Leadman5555 (changes requested):
Some minor changes to make. Also, are you 100% sure your app works correctly? Snippet from CI below:
image
Run the app locally and try to post to an endpoint, see if it is bundled correctly. The test failure message does not stem from the test itself failing, but from the bundling failure. Fix it.

DominikDab341:
I fixed all the requested changes and tested my app locally — it works fine. I'm not sure how to fix the bundling issue.
obraz_2025-09-03_151933612

Leadman5555:
Are you sure you pushed the changes to the remote? I think you might have missed that step.

DominikDab341:
Now i pushed... Sorry about that

src/trip/trip.service.ts
```diff
@@ -0,0 +1,85 @@
...
+  async update(id: number, updateTripDto: UpdateTripDto) {
+    await this.getTripOrThrow(id);
+
+    return this.database.trip.update({
+      where: { trip_id: id },
+      data: {
+        name: updateTripDto.name,
+        destination: updateTripDto.destination,
+        start_date: updateTripDto.start_date,
+        end_date: updateTripDto.end_date,
```
Leadman5555, line 73: Here is a small problem. If I wanted to remove the end date (as it is optional), can I do that in the current implementation? If I pass a value, it will get replaced with the new value. If I pass undefined or don't pass it at all (fallbacks to undefined), no changes will be made. So I effectively cannot remove the end date once it is set. But I won't make you change it since it's just a practice task - no need to make it perfect.

src/trip/trip.service.ts
```diff
@@ -0,0 +1,85 @@
...
+        budget: updateTripDto.budget,
+      },
+    });
+  }
+
+  async remove(id: number): Promise<void> {
+    await this.getTripOrThrow(id);
+    await this.database.expense.deleteMany({ where: { trip_id: id } });
```
Leadman5555, line 81: Small disclaimer, not an error: If you want cascade behaviour, you can just mark the relationship as a cascade one in the database. That way, if you delete a trip, the database engine will delete all related records automatically for you.

Leadman5555 (approved):
You can merge it if you want to. The bundling errors can be fixed in the 5th task - tests. I left two comments, but it's just for your information - no need to fix them.

PR #182, opened 2025-09-06
Zadanie 4 - Jakub Morawiec

jmorawiec:
Hej, walidacja poszła w miarę sprawnie dzięki przykładowym code snippetom, które udostępniłeś na wykładzie.

A, i gdybyś pytał o te dwa komentarze z LLMa które są w historii z commitów - posiłkuję się nim w ostateczności jak nie wiem co nie działa i jak to naprawić, tak jak było z checkiem bazy danych w GitHub Actions. Staram się nie nadużywać dobroci modeli językowych, żeby wynieść jak najwięcej z samego kursu.

Miłego sprawdzania i dzięki wielkie z góry!


src/auth/jwt.strategy.ts
```diff
@@ -0,0 +1,34 @@
+import { ExtractJwt, Strategy } from "passport-jwt";
+
+import { Injectable } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+import { PassportStrategy } from "@nestjs/passport";
+
+export interface JwtPayload {
+  sub: number;
+  email: string;
+  role: "USER" | "ADMIN";
```
Leadman5555: Use the role enum here

src/expenses/expenses.service.ts
```diff
@@ -8,15 +10,15 @@ import { UpdateExpenseDto } from "./dto/update-expense.dto";
 export class ExpensesService {
   constructor(private prisma: PrismaService) {}
 
-  async create(createExpenseDto: CreateExpenseDto) {
+  async create(createExpenseDto: CreateExpenseDto): Promise<Expense> {
     return this.prisma.expense.create({ data: createExpenseDto });
   }
 
-  async findAll() {
+  async findAll(): Promise<Expense[]> {
```
Leadman5555: This is from the last PR. Remove the expense entity. If you want types, create a response DTO and map return values to it. Exposing models like this is not the best practice because many frameworks may serialize additional information you don't want to be exposed.

src/common/guards/trip-coordinator-or-admin.guard.ts
```diff
@@ -0,0 +1,55 @@
...
+  ForbiddenException,
+  Injectable,
+  NotFoundException,
+} from "@nestjs/common";
+
+import { PrismaService } from "../../../prisma/prisma.service";
+import type { JwtPayload } from "../../auth/jwt.strategy";
+
+@Injectable()
+export class TripCoordinatorOrAdminGuard implements CanActivate {
```
Leadman5555, line 34: Overall, I thought this task would be done a bit differently, with the COORDINATOR role being a third UserRole (together with Admin and normal User). That way doing the checks would be far easier. If you wish to keep the current implementation, it should not be a Guard. Why? 
One, it's far too error prone, especially the part with extracting the tripId.
Two, it's not really effiecient. You need to make additional DB queries here, which could be avoided if you had the proper context.
So, I see two solutions here. First option is to change the database logic. Second option is to keep the current logic and move the checks to the `tripService`. Maybe a method that is called whenever needed, that takes a tripId parameter and a User object? It could throw on failure and do nothing on success

src/common/guards/trip-coordinator-or-admin.guard.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const parameters = request.params as Record<string, string | undefined>;
+    const body = request.body as Record<string, unknown>;
+    const raw =
+      parameters.id ??
+      parameters.tripId ??
+      (body.tripId as string | number | undefined);
+    const tripId = typeof raw === "string" ? Number(raw) : Number(raw);
+
+    if (!Number.isInteger(tripId)) {
```
Leadman5555:
```ts
const tripId = typeof raw === "string" ? Number(raw) : Number(raw);
if (!Number.isInteger(tripId)) {
  ...
}
```
This sequence makes no sense.

src/common/guards/jwt-auth.guard.ts
```diff
@@ -0,0 +1,29 @@
+import { Observable } from "rxjs";
+
+import { ExecutionContext, Injectable } from "@nestjs/common";
+import { Reflector } from "@nestjs/core";
+import { AuthGuard } from "@nestjs/passport";
+
+import { IS_PUBLIC_KEY } from "../decorators/public.decorator";
+
+@Injectable()
+export class JwtAuthGuard extends AuthGuard("jwt") {
```
Leadman5555: We did not cover Passport library on the course, using it was neither obligatory, nor recommended. After all, the point is to work how it works under the hood in a simplified, minimal version. Of course, it normal application, unless you have special requirements, using a standardized library is the correct choice. Here, however, it is a bit counter-productive as it does everything for you. Especially that if you don't know how it works, you don't know how to configure a ready-made solution either. 
Nevertheless, I'll accept the guard solution without checking it in detail.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const payload = {
+      sub: user.id,
+      email: user.email,
+      role: user.role,
+      iat: nowSec,
+      exp: expSec,
+    };
+    const token = this.jwt.sign(payload, {
+      secret: this.config.get<string>("JWT_SECRET") ?? "",
```
Leadman5555: The secret can probably be injected into the Passport module declaration. Check if it is possible. Especially that you do pass it in the `jwt.strategy.ts`. I don't know the passport library, so I'm not too sure, but passing the token here may override the previous value. Check it and determine whether you need to pass the token here, in the strategy or in module declaration.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,69 @@
...
+      this.config.get<string>("EXPIRY_TIME_MS") ?? 3_600_000,
+    );
+    const expSec = nowSec + Math.floor(expMs / 1000);
+
+    const payload = {
+      sub: user.id,
+      email: user.email,
+      role: user.role,
+      iat: nowSec,
+      exp: expSec,
```
Leadman5555: Again, I don't know the passport library, but I'm pretty sure you should not pass subject and expiry in directly in the payload. They probably go into the sign options. That's why using libraries without knowing the basics may lead to errors you will never know about.

src/users/users.service.ts
```diff
@@ -0,0 +1,57 @@
...
+@Injectable()
+export class UsersService {
+  constructor(private prisma: PrismaService) {}
+
+  async findById(id: number) {
+    const user = await this.prisma.user.findUnique({ where: { id } });
+    if (user === null) {
+      throw new NotFoundException("Użytkownik nie znaleziony");
+    }
+    const { passwordHash, ...safe } = user;
```
Leadman5555, line 22: Create a response DTO with a mapping function instead

src/users/users.service.ts
```diff
@@ -0,0 +1,57 @@
...
+      role?: "USER" | "ADMIN";
+    } = {};
+
+    if (typeof dto.email === "string" && dto.email.length > 0) {
+      data.email = dto.email;
+    }
+    if (typeof dto.password === "string" && dto.password.length > 0) {
+      data.passwordHash = await bcrypt.hash(dto.password, 12);
+    }
+    if (dto.role === "USER" || dto.role === "ADMIN") {
```
Leadman5555, line 46: This check is pointless - the role is always one of the two.

Leadman5555 (changes requested):
There are some changes to be made. Also, I don't think you have included any custom validators - you need one to meet the requirements.

jmorawiec:
Hi @Leadman5555 , thank you for your detailed review! I will apply the changes on Monday as I have the day off today.

src/auth/auth.service.ts
jmorawiec: Słuszna uwaga, za mało dokumentacji przeczytałem, przepisałem z GitHuba to co znalazłem i tak to się skończyło

jmorawiec:
All the changes have been applied, @Leadman5555 please re-check. Thank you!

src/common/guards/trip-coordinator-or-admin.guard.ts
```diff
@@ -0,0 +1,75 @@
...
+  body: ParameterBag,
+): number | null {
+  const candidate = parameters.tripId ?? parameters.id ?? body.tripId;
+
+  if (typeof candidate === "number") {
+    return Number.isInteger(candidate) && candidate > 0 ? candidate : null;
+  }
+  if (typeof candidate === "string") {
+    // nie umiem regexów 😭 ale idk czy da się inaczej
+    return /^[1-9]\d*$/.test(candidate) ? Number(candidate) : null;
```
Leadman5555, line 28: Overall, using regex is usually not the best call if it can be avoided. Typescript has a nice method in the Number namespace, called `parse` I think. And another method called `isNaN` which basically checks if the parsing went correctly or not.

src/common/guards/trip-coordinator-or-admin.guard.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return null;
+}
+
+@Injectable()
+export class TripCoordinatorOrAdminGuard implements CanActivate {
```
Leadman5555, line 34: I think you might have... well, completely misunderstood my comment about this guard. But no matter, we are already behind the schedule and not so many people are left, so it's not that important.

Leadman5555 (approved):
I'll accept the PR.  There still are some minor issue I would usually address, but the projects are coming and they are far more important. I've also realized that I might have been a little to strict with reviews in my group...

PR #193, opened 2025-09-09
feat/ Dominik Kaczmarek 5

feen1e:
Solution for task 5


src/participant/participant.controller.spec.ts
AlicjaBonar, line 1: You could simplify the tests by introducing a small helper that returns a default UserMetadata object instead of repeating the same user creation in multiple places. This would reduce duplication, make the tests easier to read, and allow you to change the test user’s details in one central location if needed.
If you don't want to do it now, then no problem, but keep it in mind for the future

AlicjaBonar (approved):
Great work! The tests have solid coverage. One suggestion for the future: consider creating helper functions for the DTOs used in the tests, as they’re repeated often. This would make the tests cleaner and easier to maintain.

PR #199, opened 2025-09-09
Dominik dabrowski 4

DominikDab341:


prisma/schema.prisma
```diff
@@ -42,6 +45,7 @@ model Participant {
   participant_id Int    @id @default(autoincrement())
   trip_id        Int
   trip           Trip   @relation(fields: [trip_id], references: [trip_id])
+  User        User     @relation(fields: [email], references: [email])
   first_name     String @db.VarChar(32)
   last_name      String @db.VarChar(32)
   role           Role
```
Leadman5555: The fields should not repeat. We have a role field on User. We have a second role field on the Participant. That means that they can have different roles. We need to fix that. Unless there is a valid reason for, if so, feel free to explain in the comments below

src/auth/auth.guard.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    if (authHeader === undefined || authHeader.length === 0) {
+      throw new UnauthorizedException("No token provided");
+    }
+
+    if (!authHeader.startsWith("Bearer ")) {
+      throw new UnauthorizedException("Invalid token format");
+    }
+
+    const token = authHeader.split(" ")[1];
```
Leadman5555: Using split here is not the best idea. There might be a token, now or in the future, that just happens to have an extra whitespace inside it. If you use split, you will only get a part of the token. It's better to get a substring instead

src/auth/auth.service.ts
```diff
@@ -0,0 +1,152 @@
...
+        name: dto.name ?? null,
+        password,
+        role: Role.USER,
+        isEnabled: true,
+      },
+    });
+  }
+
+  async validateToken(token: string): Promise<UserMetadata | null> {
+    if (!token || typeof token !== "string") {
```
Leadman5555: If you declare token as string, it means you know it is a string. So if you know, then why check it? If you don't know if it's a string, then declare it as unknown

Leadman5555: The token comes from the headers. See what type do they return and change this method accordingly

src/auth/auth.service.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const parts = token.slice(this.tokenPrefix.length).split("_");
+    if (parts.length !== 2) {
+      return null;
+    }
+
+    const timestamp = Number.parseInt(parts[0], 10);
+    const email = parts[1];
+
+    if (Number.isNaN(timestamp) || !email) {
```
Leadman5555: Don't use shorthand checks like for email. It's bad practice and if you ever refactor the code and change things, you might miss such checks and have unexpected behaviour.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,152 @@
...
+      return null;
+    }
+
+    const expiryTimeMs = Number.parseInt(
```
Leadman5555: Extract the expiry time to a class field. There is no point to fetch it and calculate it with every single validated request

src/auth/auth.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    ) {
+      throw new UnauthorizedException();
+    }
+    return { token: this.generateToken(user.email) };
+  }
+}
+
+// @Injectable()
+// export class AuthService {
+//   private readonly tokenPrefix = "token_";
```
Leadman5555: I guess you missed that comment out code. Let's remove it

src/auth/dto/login.dto.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ApiProperty } from "@nestjs/swagger";
+
+export class LoginDto {
+  @ApiProperty({
+    description: "User email address",
+    example: "user@example.com",
+  })
+  @IsEmail({}, { message: "Please provide a valid email address" })
+  @Transform(({ value }: { value: unknown }) =>
+    typeof value === "string" ? value.toLowerCase().trim() : value,
```
Leadman5555: This transformer is redundant. If the value passes the email check, then it must be a string. Furthermore, in your transformer, if the value is not a string, you keep it. And later it is assigned to a string field anyway.

src/auth/dto/register.dto.ts
```diff
@@ -0,0 +1,49 @@
...
+import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
+
+export class RegisterDto {
+  @ApiProperty({
+    description: "User email address",
+    example: "user@example.com",
+  })
+  @IsEmail({}, { message: "Please provide a valid email address" })
+  @Transform(({ value }: { value: unknown }) =>
+    typeof value === "string" ? value.toLowerCase().trim() : value,
```
Leadman5555: Same thing here

src/auth/dto/register.dto.ts
```diff
@@ -0,0 +1,49 @@
...
+      "Password must contain at least one uppercase letter, one lowercase letter, and one number",
+  })
+  password: string;
+
+  @ApiPropertyOptional({
+    description: "User full name",
+    example: "John Doe",
+  })
+  @IsOptional()
+  @IsString({ message: "Name must be a string" })
```
Leadman5555: There is actually no need to provide such messages. The autogenerated error message will give the reason (wrong type) and the field

src/auth/dto/register.dto.ts
```diff
@@ -0,0 +1,49 @@
...
+      "User password (min 8 chars, must contain uppercase, lowercase and number)",
+    example: "Password123",
+  })
+  @IsString({ message: "Password must be a string" })
+  @MinLength(8, { message: "Password must be at least 8 characters long" })
+  @Matches(/^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)/, {
+    message:
+      "Password must contain at least one uppercase letter, one lowercase letter, and one number",
+  })
+  password: string;
```
Leadman5555, line 35: It's good to add a max length here and on email

src/auth/dto/register.dto.ts
```diff
@@ -0,0 +1,49 @@
...
+  @ApiPropertyOptional({
+    description: "User full name",
+    example: "John Doe",
+  })
+  @IsOptional()
+  @IsString({ message: "Name must be a string" })
+  @MinLength(2, { message: "Name must be at least 2 characters long" })
+  @MaxLength(100, { message: "Name cannot exceed 100 characters" })
+  @Transform(({ value }: { value: unknown }) =>
+    typeof value === "string" ? value.trim() : value,
```
Leadman5555: Again, this transformer is redundant

src/auth/dto/register.dto.ts
```diff
@@ -0,0 +1,49 @@
...
+  )
+  name?: string | null;
```
Leadman5555: This will lead to errors. You have a `@IsString` validator and an optional validator. What will happen if the name is null?

src/participant/dto/create-participant.dto.ts
```diff
@@ -1,44 +1,71 @@
...
+    minLength: 2,
+    maxLength: 50,
   })
-  @IsString()
-  @MinLength(2)
+  @IsString({ message: "First name must be a string" })
+  @MinLength(2, { message: "First name must be at least 2 characters long" })
+  @MaxLength(50, { message: "First name cannot exceed 50 characters" })
+  @Transform(({ value }: { value: unknown }) =>
+    typeof value === "string" ? value.trim() : value,
```
Leadman5555: I won't repeat it anymore, but remove the transform validators or change them

src/participant/participant.service.ts
```diff
@@ -65,7 +65,9 @@ export class ParticipantService {
       first_name: updateDto.first_name,
       last_name: updateDto.last_name,
       role: updateDto.role,
-      email: updateDto.email,
+      ...(updateDto.email != null && {
+        User: { connect: { email: updateDto.email } },
```
Leadman5555: You need to check first if the user exists. The same with trip below. Remember about it in the other endpoints

src/trip/dto/trip-private-response.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  @ApiProperty({ example: "2025-09-01" })
+  start_date: Date;
+
+  @ApiPropertyOptional({ example: "2025-09-03" })
+  end_date?: Date;
+
+  @ApiPropertyOptional({ example: 1200 })
+  budget?: number;
+
+  @ApiProperty({ type: [ParticipantResponseDto] })
```
Leadman5555: This returns an array, not a single DTO

src/user/user.controller.ts
```diff
@@ -0,0 +1,143 @@
...
+  @ApiResponse({
+    status: 403,
+    description: "Forbidden - can only view own data",
+  })
+  @ApiResponse({ status: 404, description: "User not found" })
+  async findOne(
+    @Param() parameters: EmailParameterDto,
+    @Request() request: AuthenticatedRequest,
+  ): Promise<User> {
+    this.checkUserAccess(request.user, parameters.email);
```
Leadman5555: Actually, do we really need this check? What if the endpoint, instead of taking an email param from the request, would take the email value from the authentication principal? That way, you would always access your own data

prisma/schema.prisma
```diff
@@ -59,4 +63,13 @@ model Expense {
   cost         Decimal     @db.Decimal(10, 2)
   description  String?     @db.VarChar(255)
 
+}
+
+model User {
+  email       String        @id
```
Leadman5555, line 68: Using an email as an Id might not be the best choice. It's usually better to leave it as a separate unique column with an index on it (if the database supports indexes, that is).
But let's leave it as it is here.

src/validators/email-parameter.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { Transform } from "class-transformer";
+import { IsEmail } from "class-validator";
+
+export class EmailParameterDto {
```
Leadman5555, line 3: Why is it wrapped in a class if it is a single value? Judging by the name of it, you do not plan to expand it in the future

src/validators/future-date.validator.ts
```diff
@@ -0,0 +1,43 @@
...
+@ValidatorConstraint({ async: false })
+export class IsFutureDateConstraint implements ValidatorConstraintInterface {
+  validate(date: unknown, _arguments: ValidationArguments): boolean {
+    if (typeof date !== "string" && !(date instanceof Date)) {
+      return false;
+    }
+
+    const inputDate = new Date(date);
+
+    if (Number.isNaN(inputDate.getTime())) {
```
Leadman5555: I assume you are checking the the validity of the date. There probably is a better way of doing so

src/validators/future-date.validator.ts
```diff
@@ -0,0 +1,43 @@
...
+      return false;
+    }
+
+    const today = new Date();
+    today.setHours(0, 0, 0, 0);
```
Leadman5555: Maybe we could just extract the date part? This method needs a bit of more thought. The current implementation might work, but it's a bit of a workaround

src/validators/id-parameter.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { Transform } from "class-transformer";
+import { IsInt, Min } from "class-validator";
+
+export class IdParameterDto {
```
Leadman5555, line 3: Same as with email parameter DTO

Leadman5555 (changes requested):
A couple of things to change, but the overall direction is good

prisma/schema.prisma
DominikDab341: A participant in one trip can be an organizer, while in another trip the same person may only be a member.
At the same time, a User in the system has a global role, which is set to USER by default, but it can be updated in the database to ADMIN if needed.

src/auth/auth.service.ts
DominikDab341: I’d prefer to keep the commented-out code for now, as it helps me quickly test API responses that are restricted to specific roles.

src/validators/email-parameter.dto.ts
DominikDab341: I decided to keep the DTO because it keeps the code consistent and self-documented, even if it’s just one field. In my opinion, this is also a relatively easier solution.

src/validators/future-date.validator.ts
```diff
@@ -0,0 +1,40 @@
...
+import type {
+  ValidationArguments,
+  ValidationOptions,
+  ValidatorConstraintInterface,
+} from "class-validator";
+
+@ValidatorConstraint({ async: false })
+export class IsFutureDateConstraint implements ValidatorConstraintInterface {
+  validate(date: unknown, _arguments: ValidationArguments): boolean {
+    if (typeof date !== "string" && !(date instanceof Date)) {
```
Leadman5555: Now that I think of it, the second check `!(date instanceof Date)` is pointless. If used in context the given value will never be a date object as we get the value from a request. Even if it were a date object (used outside of context), it might still fail a few lines down when you cast it as string.

The previous implementation was better then the new one and without using luxon, it is probably the simplest one. You can use something like this:
```ts
if (typeof givenValue !== "string") {
    return false;
}
const then = new Date(givenValue );
if(Number.isNaN(then.valueOf())) {
    return false;
}
const now = new Date();
now.setHours(0, 0, 0, 0);
then.setHours(0, 0, 0, 0);
return now.valueOf() >= then.valueOf();
 ```

src/validators/email-parameter.dto.ts
Leadman5555: Alright, I can accept that explanation 👍

prisma/schema.prisma
Leadman5555: Okay. But if so, we should still make a distinction between the two roles. The User role is connected to management overall, while the Participant role is tied to a single trip instance. Let's make a separate enum called TripRole or whatever you want to. Move the trip connected values to it.

src/auth/auth.service.ts
Leadman5555: Well, it has no place in the final PR. After all, we don't want anything like this in our codebase. My advice is to either remove it before you send the PR and simple paste it in from notepad whenever you need to (which is what I sometimes personally do) or to create a separate testing controller and add it to gitignore so it does not appear on GitHub - that way you can also test things without bothering about the linter in CI. Either way, it cannot appear in the final PR

Leadman5555 (changes requested):
A few minor changes to make, but nearly there.

Leadman5555 (approved):
LGTM 👍

PR #203, opened 2025-09-11
Dawid Błaszczyk - Zadanie 5

Skiperpol:


src/expense/expense.controller.spec.ts
AlicjaBonar, line 1: You can make the tests shorter by adding a simple helper that builds expense data for you.
This way you don’t have to write the same expense objects again and again, and it’s easier to update them later.

src/expense/expense.service.spec.ts
AlicjaBonar, line 1: Same case as with expense controller, you could add little helpers for example data. This applies to other tests as well

test/expenses.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  beforeAll(async () => {
+    await cleanDatabase();
+    await seedDatabase();
+  });
```
AlicjaBonar, line 30: Your current approach can leave the database in a different state after each test—for example, a test might insert or delete records.
If one test fails and wipes data, the next tests could also fail when they try to update or read records that no longer exist.
To avoid this, it’s better to reset or seed the database in a beforeEach so every test starts with a clean, consistent state.

AlicjaBonar (approved):
Great work! Just two small insights

PR #209, opened 2025-09-12
Hubert drewlo - task 4

Drewienkopl:
I made it so all GET, register and login are public, POST and PATCH aren't, and trips POST and PATCH are only for admins/cordinators.

This is result of this task: 

You can GET without login:
1_expenceGetWorks

You can't POST unauthorised:
2_expencePostWorksNot

Registering an account works:
_register1Works

You can't create second account with the same email, it gives conflict error:
_register1NotWorksConflict

You can login:
_login1Works

Wrong password = you can't login and give "Incorrect password" message with unauthorized error:
_login1WorksNOT

Inputting not existing user email gives "User not found" message:
_login2WorksNOT

After logging in using token, now it's possible to POST:
POST_WorksWhenLoggedIn

Putting in wrong token gives "Invalid token message":
InvalidToken

Changing your own data works:
ChangingYourNameWorks

You can't change another account's data:
OnlyYourOwn

Users can't change role:
image

Validator works:
ValidatorWorks

Admin/Cordinator can create new trips:
adminCan

Users can't create new trips:
UsersCant


src/auth/auth.service.ts
```diff
@@ -0,0 +1,57 @@
...
+
+@Injectable()
+export class AuthService {
+  private readonly tokenPrefix = "token_";
+
+  constructor(private userService: UserService) {}
+
+  generateToken(email: string): string {
+    const ts = Date.now();
+    const payload = `${email}::${String(ts)}`;
```
Leadman5555: It's better to use `ts.toString()` instead

src/auth/auth.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    } catch {
+      throw new Error("Invalid token");
+    }
+    const [email, tsString] = payload.split("::");
+    const ts = Number(tsString);
+    if (!email || Number.isNaN(ts)) {
+      throw new Error("Invalid token");
+    }
+
+    const expiryMs = Number(process.env.EXPIRY_TIME_MS ?? "86400000");
```
Leadman5555: Move this to a class field. Initialize it once in the constructor. There is no point in reading the env file with every single authenticated request

src/auth/dto/register.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  @ApiProperty({ minLength: 8, writeOnly: true })
+  @IsString()
+  @MinLength(8)
+  password: string;
+
+  @ApiPropertyOptional({ example: "Joe" })
+  @IsOptional()
+  @IsString()
+  name?: string;
```
Leadman5555, line 27: It's good to add max length constraints on every property. Otherwise, we risk someone putting a **very** long string and potentially causing 500 errors, depending on how the schema is defined.

src/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,61 @@
...
+      request.headers.authorization ?? request.headers.Authorization;
+    const header: string | undefined = Array.isArray(rawHeader)
+      ? rawHeader[0]
+      : rawHeader;
+
+    if (typeof header !== "string") {
+      throw new UnauthorizedException("Missing Authorization header");
+    }
+
+    const token = header.startsWith("Bearer ") ? header.slice(7) : header;
```
Leadman5555: This check is not need. If the header does not start with "Bearer " simply throw 401. There are many authentication methods and they all follow certain guidelines. Bearer tokes are always in a bearer header.

src/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,61 @@
...
+      return true;
+    }
+
+    const request = context
+      .switchToHttp()
+      .getRequest<Request & { user?: unknown }>();
+
+    const rawHeader =
+      request.headers.authorization ?? request.headers.Authorization;
+    const header: string | undefined = Array.isArray(rawHeader)
```
Leadman5555: Auth header should never be an array, so we don't need this extraction here. Even if it somehow were, than that just means the request is invalid and we should not bother checking it. We expect one schema and, if anyone wants to get authenticated, they are expected to follow that schema. 
Also, some frameworks outright reject requests if there are multiple auth headers

src/auth/guards/roles.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (requiredRoles == null || requiredRoles.length === 0) {
+      return true;
+    }
+
+    const request = context.switchToHttp().getRequest<{ user?: RequestUser }>();
+
+    const user = request.user;
+
+    if (!user?.role) {
+      throw new ForbiddenException("No roles assigned");
```
Leadman5555: That's not really correct. In your schema, the role is a required field. So a user fetched from database always has a role. Here, the lack of role, or rather, the user, can only happen if role guard was executed without the auth guard before it. That's a error on the side of the software developer. I'd throw a code in the 500s. Maybe normal 500 or 503

src/user/user.controller.ts
```diff
@@ -0,0 +1,68 @@
...
+  @ApiForbiddenResponse({
+    description: "You are not allowed to update this user",
+  })
+  @ApiBadRequestResponse({ description: "Validation failed" })
+  async updateUser(
+    @Param("email") email: string,
+    @Body() dto: UpdateUserDto,
+    @CurrentUser() currentUser: { email: string; role: string },
+  ) {
+    if (currentUser.role === "ADMIN") {
```
Leadman5555: Use the enum here instead of a magic string

src/user/user.controller.ts
```diff
@@ -0,0 +1,68 @@
...
+      return this.userService.updateUser(email, dto);
+    }
+
+    if (currentUser.email !== email) {
+      throw new ForbiddenException("You can only modify your own data");
+    }
+
+    if (dto.role !== undefined && dto.role !== currentUser.role) {
```
Leadman5555, line 64: It's usually better to have a separate endpoint for role modifications and such. But let's leave it as it is 👍

src/user/user.service.ts
```diff
@@ -0,0 +1,116 @@
...
+  async findByEmailOrFail(email: string) {
+    const user = await this.prisma.user.findUnique({ where: { email } });
+    if (user === null) {
+      throw new NotFoundException("User not found");
+    }
+    return user;
+  }
+
+  async findMetadataOrFail(email: string) {
+    const u = await this.prisma.user.findUnique({
```
Leadman5555: Extract the flow "find one or throw" into a single method. Or you can use the `findByEmailOrFail` method for that.

src/user/user.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      where: { email },
+      select: { email: true, role: true, isEnabled: true },
+    });
+    if (u === null) {
+      throw new NotFoundException("User not Found");
+    }
+    if (!u.isEnabled) {
+      throw new NotFoundException("User disabled");
```
Leadman5555: This should not be 404, it should be 401 or 403

src/user/user.service.ts
```diff
@@ -0,0 +1,116 @@
...
+        email: true,
+        name: true,
+        role: true,
+      },
+    });
+  }
+
+  async updateUser(
+    email: string,
+    dto: Partial<{
```
Leadman5555: Declare a normal DTO with validators for this

src/user/user.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      name?: string | null;
+      role?: Role;
+      additionalInfo?: string | null;
+    } = {};
+
+    if (dto.email !== undefined) {
+      data.email = dto.email;
+    }
+
+    if (dto.password !== undefined && dto.password !== "") {
```
Leadman5555: We can remove checks like `dto.password !== ""` if we use a declared DTO with validators

Leadman5555 (changes requested):
Some minor changes to be made

Leadman5555:
Take note of the commit names or otherwise the CI will fail. What caused the error here, was "Feat: Task...". It expects the prefix to be lowercase ("feat" instead of "Feat") and the name to start with a lowercase letter ("task" instead of "Task"). 
Personally, I don't really like the naming schema, especially the part with the name being lowercase, but since someone link the naming schema here, we should use it regardless.

Leadman5555 (approved):

PR #214, opened 2025-09-15
feat/ Dominik Kaczmarek 6

feen1e:
Added solution for list 6: implemented scheduled scraping of currency exchange rates and created a payments module.

Thanks for always giving helpful tips and comments over the past few weeks! I had fun and learnt a lot :)


src/currencies/currency-scraper.service.ts
AlicjaBonar, line 1: You are missing error catching in the scarpper
It would be better to do something like this:

try {
  // scraping
} catch (e) {
  this.logger.error(e);
  throw e;
} finally {
  await browser.close();
}

Instead of commenting out mamy different functionalities you could change them into separate functions that could be called out as needed

AlicjaBonar (approved):
Great job! Just a small suggestion for improvement.
Thank you for participating in this course, it’s been a pleasure to see your code evolve. 😊

PR #215, opened 2025-09-16
Hubert drewlo - task 5

Drewienkopl:
Task 5 result. I have managed to fix most of the errors i was getting and when running tests one by one they all pass
image

But when running them all together a few always fail:
image

I divided tests into 3 since when they run all at once a few would always fail.

I added some more e2e tests.


docker-compose.test.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.8"
+services:
```
Leadman5555: We don't need docker here. Actually, nothing what we have here has anything in common with it. But I do suspect why you put it here - I'll explain in review comment.
You can remove this file.

src/validators/is-future-date.validator.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { validate } from "class-validator";
+
+import { IsFutureDate } from "./is-future-date.validator";
+
+class Dto {
+  @IsFutureDate({ message: "must be future" })
+  date?: string | null;
```
Leadman5555: Why do we have the undefined modifier `?` here?

test/utils/register-login.ts
```diff
@@ -0,0 +1,51 @@
+import { Role } from "@prisma/client";
+import type { Server } from "node:http";
+import request from "supertest";
+
+import type { LoginResponseDto } from "../../src/auth/dto/login-response.dto";
+import { app, prisma } from "../setup";
+
+export async function registerAndLogin(
```
Leadman5555: I actually think this utils suite is a bad idea unless used in a system test.
Why? e2e tests are also known as integration tests. They test how components that form one aspect of the application work together. For example, participants IT (persistence layer, service, controller). They should, however, test only one aspect unless impossible otherwise - too closely connected for example. 
Tests that deal with a larger scope or the whole application are called system tests, often done as live tests.
Here, in each e2e test, apart from testing that one aspect, you also test authorization. And that's not really correct.

It would be easier and better if instead, before each test case that requires real database, you removed all data (as you do now) and added sample one. For example, one login-available user. Then, in tests, you do not create user and login as one. You use the predefined user and a predefined valid authentication token. 
TLDR: add sample data to the db, reset it before each test case. Use the sample data with predefined auth tokens

Leadman5555 (changes requested):
I haven't checked the e2e tests yet, I will check them after CI is fixed.
The reason why the CI fails, is because, as you can see in workflow logs: (https://github.com/Solvro/backend-wakacyjne-wyzwanie-2025-zadania/actions/runs/17769831200/job/50502652691), there is no database connection. 

You tried setting up a docker compose but docker is not used during the CI. Also, the compose file is not correct, even if it were.

So, the fix... you need to add database to the CI file (in `workflows/ci.yml`), Actually, another participant in the course had the same problem, see: https://discord.com/channels/1391545238476292196/1415693554155524176 (Q&A thread on our discord server)

In the future, if you have any problems that you cannot think of a way to fix, feel free to ask on the Q&A channel or message me directly. That's the reason I am here, after all 👍

test/expense.e2e-spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        lastname: "Payer",
+        email: "payer@example.com",
+        tripId,
+      },
+    });
+    participantId = participant.id;
+  });
+
+  afterAll(async () => {
+    await prisma.expense.deleteMany({});
```
Leadman5555: Use TRUNCATE instead. You can move that to the helpers suite

test/participant.e2e-spec.ts
```diff
@@ -0,0 +1,117 @@
...
+      .expect(201);
+
+    expect(response.body).toMatchObject({
+      name: "John",
+      lastname: "Doe",
+      tripId,
+    });
+  });
+
+  it("PATCH /participants/:id should update when authenticated", async () => {
```
Leadman5555, line 81: Test what happens if the trip of the given Id does not exist

Leadman5555 (changes requested):
Some tests are failing. Check workflow logs for details

Leadman5555 (changes requested):
There are still no tests for the following scenario:
A participant is created/updated and the given trip Id does not have a corresponding trip record in the database

You add a test for when the participant queried does not exist, but that's not all we need.

Leadman5555 (approved):

PR #219, opened 2025-09-16
Nikita Gwardiak 6

Gwarnickn:


firstassignment/src/services/currency.service.ts
```diff
@@ -0,0 +1,91 @@
+import { ConflictException, Injectable, NotFoundException } from "@nestjs/common";
+import { PrismaService } from "./prisma.service";
+import { CreateCurrencyDto } from "src/Dto/create-currency-dto";
+import * as cheerio from 'cheerio';
+import { Cron, CronExpression } from "@nestjs/schedule";
+
+
+@Injectable()
+export class CurrnecyService{
```
AlicjaBonar, line 9: Small typo in Currnecy istead of currency

firstassignment/src/services/currency.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    async currencyByName(name: string){
+        return this.prisma.currency.findUnique({where: {name}});
+    }
+    
+    async allCurrencies(){
+        return this.prisma.currency.findMany();
+    }
+
+    async createCurrency(data: CreateCurrencyDto){
+        const currency = await this.currencyByName("");
```
AlicjaBonar, line 25: Here you search for currency “” instead of data.name, so it will create errors

firstassignment/src/services/currency.service.ts
```diff
@@ -0,0 +1,91 @@
...
+            {
+                name: "#js-currency-rates-container > table > tbody > tr.row-6 > td.rate-border-right.uk-text-bold.uk-text-center.uk-text-primary",
+                rate: "#js-currency-rates-container > table > tbody > tr.row-6 > td:nth-child(3)",
+            }
+        ]
+
+        try{
+            const response = await fetch('https://www.bankbps.pl/kursy-walut?view=currencies');
+            if(!response.ok){
+                throw new Error(`Błąd HTTP! status: ${response.status.toString()}`)
```
AlicjaBonar, line 73: I don’t think you need to do toString() here, it’s not important, but I’d rather point it out then not

firstassignment/src/services/currency.service.ts
```diff
@@ -0,0 +1,91 @@
...
+                data: {
+                    rate
+                },
+                where: {name}
+            })
+        }
+    }
+
+    @Cron(CronExpression.EVERY_2_HOURS)
+    async downloadCurrencies(){
```
AlicjaBonar, line 54: I noticed the function doesn’t return a value. That’s not a bug, but returning something, like the number of currencies downloaded, would make later testing and verification easier.

AlicjaBonar (approved):
Great job! Just a small suggestion for improvement and one little thing that actually might cause errors so it will be better to correct it.
Thank you for participating in this course, it’s been a pleasure to see your code evolve. 😊

Gwarnickn:
Dzięki, Alicja za najlepsze code review, jakie miałem. Super robota i dużo z tego wyciągnąłem.

PR #222, opened 2025-09-17
Zadanie 5 - Jakub Morawiec

jmorawiec:
Wybacz że tak późno, trochę rzeczy się w moim życiu działo. Wykluczyłem testy ze sprawdzania w eslincie, bo to nie jest biblioteka którą ma zobaczyć użytkownik kodu. Jeśli się mylę i jednak powinienem checkować testy eslintem, to oczywiście daj znać. Dzięki, miłego sprawdzania!

Update: Jednak zaadresowałem wszystkie poprawki, aczkolwiek automatyczne usuwanie undefined i pustych returnów przez ESLINTa było tragiczne, a często nie wiedziałem, jak to inaczej zrobić.


src/common/guards/__tests__/roles.guard.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    getHandler: jest.fn(),
+    getClass: jest.fn(),
+  } as unknown as ExecutionContext;
+}
+
+describe("RolesGuard", () => {
+  it("odrzuca gdy brak usera", () => {
+    const reflector = {
+      getAllAndOverride: jest.fn().mockReturnValue(["ADMIN"]),
+    } as unknown as Reflector;
```
Leadman5555, line 25: Actually, you can use a real reflector instance here and just mock the return value of a particular method. Nevertheless, he current solutions works 👍

src/common/validators/__tests__/is-after.validator.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { validate } from "class-validator";
+import "reflect-metadata";
+
+import { IsAfter } from "../is-after-validator";
+
+class DummyDto {
+  startDate!: string;
```
Leadman5555: There is no need to use the non-null type hint here - mainly because linter usually screams at you if you do. You can give the field a default value, for example "2025-01-01T10:00:00Z" and just set the end date in tests. If needed, you can change the fields value to any other later as well.

src/participants/__tests__/participants.service.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+        { provide: PrismaService, useValue: prismaMock },
+      ],
+    }).compile();
+
+    service = moduleRef.get(ParticipantsService);
+  });
+
+  afterEach(() => jest.clearAllMocks());
+
+  it("create() tworzy uczestnika", async () => {
```
Leadman5555, line 43: Create a test that checks what happens if the given trip does not exist

test/e2e/participants.e2e-spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    prisma = new PrismaClient();
+    app = await buildParticipantsApp();
+    server = app.getHttpServer() as http.Server;
+  });
+
+  beforeEach(async () => {
+    await resetDatabase();
+  });
+
+  afterAll(async () => {
```
Leadman5555, line 25: That's not really need as it will happen by itself when the variables go out of scope. Although having it here is not a mistake - it just means that it won't be called later by the framework

test/e2e/participants.e2e-spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    await prisma.$disconnect();
+    await app.close();
+  });
+
+  it("POST /participants -> 201", async () => {
```
Leadman5555, line 30: Check what happens if the trip does not exist

test/utils/test-app.factory.ts
```diff
@@ -0,0 +1,47 @@
...
+    .compile();
+
+  const app = moduleRef.createNestApplication();
+  await app.init();
+  return app;
+}
+
+export async function resetDatabase(): Promise<void> {
+  const prisma = new PrismaClient();
+  await prisma.expense.deleteMany().catch(() => null);
```
Leadman5555: This will not reset everything properly. It depends on how you defined the primary keys - since you have autoincrement in yours, it maps to a sequence type in the database. Serial types do not get reset when records are dropped. It just didn't affect you in your tests, because you always create records as needed.
You should either drop and reset the serial which is `TRUNCATE` or you should refresh the whole schema. Truncating would be better here

Leadman5555: In many test suites, you usually use a predefined seeded database which is the same for all e2e tests with the exact same records. If you try to renew it after each test case, that's where this will matter.

Leadman5555 (changes requested):
A few minor things to fix

Leadman5555 (approved):

PR #224, opened 2025-09-17
Jakub ptak 4

ptakj:


src/auth/auth.controller.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+  @ApiResponse({
+    status: 200,
+    description: "Registered",
+  })
+  @ApiResponse({
+    status: 401,
+    description: "No permision or account disabled",
+  })
+  @HttpCode(HttpStatus.OK)
```
Leadman5555: This should be 201 CREATED

src/auth/auth.guard.ts
```diff
@@ -0,0 +1,32 @@
...
+import { LoginResponseDto } from "./dto/response-login.dto";
+
+@Injectable()
+export class AuthGuard implements CanActivate {
+  constructor(private authService: AuthService) {}
+
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const request: LoginResponseDto = context.switchToHttp().getRequest();
+
+    const [type, token] = request.headers.authorization?.split(" ") ?? [];
```
Leadman5555: Splitting by whitespace is not the best idea. If the token had any whitespaces in it (although it should not, but who know?) you will not extract the full token. Use `substring` instead

src/auth/auth.guard.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (type !== "Bearer") {
+      throw new UnauthorizedException("Missing token");
+    }
+
+    try {
+      await this.authService.validateToken(token);
+    } catch (error) {
```
Leadman5555: I don't really see the point of this try/catch block. Such blocks on auth flows are mainly used to hide what went wrong and increase security. If you rethrow with an error message, it's pointless. Either send a default error message or don't catch the errors at all

src/auth/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+import { UserMetadata, userToMetadata } from "../user/dto/user-metadata";
+import { UserService } from "../user/user.service";
+
+@Injectable()
+export class AuthService {
+  private readonly tokenPrefix = "TOKEN";
+
+  constructor(private userService: UserService) {}
+
+  private expiryTime = 2000000000;
```
Leadman5555: It's good to add a unit of measurement here in the variable name

src/auth/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    } catch {
+      userExists = true;
+    }
+
+    if (userExists) {
+      throw new ConflictException("User already exists");
+    }
+
+    const salt = 10;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
```
Leadman5555: We shouldn't do that. It is 100% possible to avoid the linter errors by specifying types correctly

src/auth/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+      throw new Error("Token expired");
+    }
+    const metaDataReturn = await this.userService.findOneOrFail(email);
+    return userToMetadata(metaDataReturn);
+  }
+
+  async signIn(email: string, password: string) {
+    const usr = await this.userService.findOneOrFail(email);
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
Leadman5555: Here as well. Also, note the password check can throw, so you should default any errors to `false`

src/auth/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const currentTime = Date.now().toString();
+
+    return { token: `${this.tokenPrefix}__${email}__${currentTime}` };
+  }
+
+  async registerAuth(registerUserDto: RegisterUserDto) {
+    let userExists = false;
+    try {
+      await this.userService.findOneOrFail(registerUserDto.email);
```
Leadman5555: I don't think this will work as it is supposed to - rethink the flow here. `findOneOrFail` throws is none is found - in this case, we want no duplicates, so if it throws, the user does not exist.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    } catch {
+      userExists = true;
+    }
+
+    if (userExists) {
+      throw new ConflictException("User already exists");
+    }
+
+    const salt = 10;
```
Leadman5555: That's not the salt in Bcrypt. This is the number of rounds.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    try {
+      await this.userService.registerUser({
+        email: registerUserDto.email,
+        login: registerUserDto.login,
+        password: hashedPassword,
+        age: registerUserDto.age,
+        description: registerUserDto.description,
+      });
+    } catch {
+      throw new InternalServerErrorException("User could not be created");
```
Leadman5555: This will throw 500 by itself in case of failure, but rethrow can be treated as error silencing to security reasons

src/trip/dto/create-trip.dto.ts
```diff
@@ -1,12 +1,16 @@
 import { ApiProperty } from "@nestjs/swagger";
+import { IsDate, IsString, MaxLength } from "class-validator";
 
 export class CreateTripDto {
+  @IsDate()
   @ApiProperty()
   startDate: Date;
```
Leadman5555, line 9: Use a transformer with date result here. This might fail on input

src/trip/dto/response-trip.dto.ts
```diff
@@ -1,12 +1,17 @@
 import { ApiProperty } from "@nestjs/swagger";
+import { IsDate, IsString, MaxLength } from "class-validator";
 
 export class ResponseTripDto {
   @ApiProperty()
+  @IsDate()
```
Leadman5555, line 7: Here as well

src/user/dto/register-user.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+  IsString,
+  MaxLength,
+  Validate,
+} from "class-validator";
+import { NiceAge } from "src/validators/nice-age";
+
+export class RegisterUserDto {
+  @IsString()
+  @ApiProperty()
+  email: string;
```
Leadman5555, line 16: This should also have a max length constraint

src/user/user.service.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  async updateUserData(
+    email: string,
+    updateUserDto: UserUpdateDto,
+  ): Promise<User> {
+    const cleanData: Partial<UserUpdateDto> = {};
+
+    for (const key of Object.keys(updateUserDto) as (keyof UserUpdateDto)[]) {
+      const value = updateUserDto[key];
+      if (value !== undefined) {
```
Leadman5555, line 90: That's not really needed I think. If you pass a property with undefined value, no changes will be made. Although I don't know prisma that well, so please test it

src/validators/nice-age.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  ValidationArguments,
+  ValidatorConstraint,
+  ValidatorConstraintInterface,
+} from "class-validator";
+
+@ValidatorConstraint({ name: "niceAge", async: false })
+export class NiceAge implements ValidatorConstraintInterface {
+  validate(age: number, _: ValidationArguments) {
+    if (age > 120 || age < 0) {
```
Leadman5555: Just as an interesting fact:
https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://en.wikipedia.org/wiki/List_of_the_verified_oldest_people&ved=2ahUKEwiVzseGteKPAxVhefEDHa9dGWoQFnoECBkQAQ&usg=AOvVaw3EPjL2hL37X41PUsfxMvvf

Leadman5555 (changes requested):
Some things to fix here

src/auth/auth.controller.ts
ptakj: Done

src/auth/auth.guard.ts
ptakj: Done

src/auth/auth.guard.ts
ptakj: Done

src/auth/auth.service.ts
ptakj: Done

src/auth/auth.service.ts
ptakj: Done

src/validators/nice-age.ts
ptakj: Done

src/user/dto/register-user.dto.ts
ptakj: Done

Leadman5555:
All errors are connected to Prisma. It was mentioned multiple times during the lectures that you should add Prisma generation to the CI

src/auth/auth.service.ts
ptakj: Done

src/auth/auth.service.ts
ptakj: Done

src/auth/auth.service.ts
ptakj: Done

src/auth/auth.service.ts
ptakj: Done

src/trip/dto/create-trip.dto.ts
ptakj: Done

src/trip/dto/response-trip.dto.ts
ptakj: Done

src/user/user.service.ts
ptakj: Well if it works I will leave it

Leadman5555 (approved):
I seen you had quite the trouble fixing the CI, but I'm glad it worked out in the end 👍

PR #237, opened 2025-09-21
Dominik dabrowski 6

DominikDab341:


src/payment/dto/create-payment.dto.ts
```diff
@@ -0,0 +1,23 @@
...
+  @Type(() => Number)
+  amount: number;
+
+  @IsString()
+  @Length(3, 3)
+  currency: string;
+
+  @IsOptional()
+  @IsString()
+  description?: string;
```
Leadman5555, line 24: Add a max length modifier here

src/payment/payment.service.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    if (currencyData === null) {
+      throw new BadRequestException(`Currency ${currency} not supported`);
+    }
+
+    let amountInPLN: number;
+    let exchangeRate: number;
+
+    if (currency.toUpperCase() === "PLN") {
```
Leadman5555: Let's lazy fetch the `currencyData` in this case. If the currency is "PLN" there is no need to make a DB query

prisma/schema.prisma
```diff
@@ -72,3 +74,21 @@ model User {
...
+  updatedAt DateTime @default(now())
+
+}
+
+model Payment {
+  id              Int      @id @default(autoincrement())
+  originalAmount  Decimal  @db.Decimal(10, 2) 
+  originalCurrency String  @db.VarChar(3)     
+  amountInPLN     Decimal  @db.Decimal(10, 2) 
+  exchangeRate    Decimal  @db.Decimal(10, 4)
```
Leadman5555: The exchange rate is not really needed as it can be derived from `originalAmount` and `amountInPLN`, but you can leave it if you want to

prisma/schema.prisma
```diff
@@ -72,3 +74,21 @@ model User {
   name        String?
   Participant Participant[]
 }
+
+model Currency {
+  id        Int      @id @default(autoincrement())
+  code      String   @unique
```
Leadman5555: Since code is unique, let's make it the primary key. There is no need for an artificial key here

prisma/schema.prisma
```diff
@@ -72,3 +74,21 @@ model User {
...
+  code      String   @unique 
+  rate      Decimal  @db.Decimal(10, 4) 
+  updatedAt DateTime @default(now())
+
+}
+
+model Payment {
+  id              Int      @id @default(autoincrement())
+  originalAmount  Decimal  @db.Decimal(10, 2) 
+  originalCurrency String  @db.VarChar(3)
```
Leadman5555: The original currency can be linked to the Currency entity. Determine if it is worth it. If not, leave it as is.

src/scrapers/currency-scraper.ts
```diff
@@ -0,0 +1,54 @@
+import { PrismaClient } from "@prisma/client";
+import { load } from "cheerio";
+import fetch from "node-fetch";
+
+const prisma = new PrismaClient();
+
+export async function scrapeNBP() {
+  const response = await fetch(
```
Leadman5555, line 10: Check the response status before proceeding further

src/scrapers/currency-scraper.ts
```diff
@@ -0,0 +1,54 @@
+import { PrismaClient } from "@prisma/client";
+import { load } from "cheerio";
+import fetch from "node-fetch";
+
+const prisma = new PrismaClient();
+
+export async function scrapeNBP() {
```
Leadman5555, line 9: Are you sure this can be scraped via a simple GET request? When I tried to curl the site, it returned an unrendered version. From what I know, correct me if I'm wrong, Cheerio will not parse the JS code

src/currency/currency.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable, Logger } from "@nestjs/common";
+import { Cron, CronExpression } from "@nestjs/schedule";
+
+import { scrapeNBP } from "../scrapers/currency-scraper";
+
+@Injectable()
+export class CurrencyService {
+  private readonly logger = new Logger(CurrencyService.name);
+
+  @Cron(CronExpression.EVERY_MINUTE)
```
Leadman5555: That's a bit too often. Change it to once an hour or every 30 minutes at least.

Leadman5555 (changes requested):
A few things to fix. Most importantly, check if the scraper code actually works - what is fetched via the `node.fetch` request. Since it behaves similar to a simple curl, it should return the same, and curl did not work as intended.

Leadman5555:
Also, some tests failed, so that's something that should be fixed as well

src/currency/currency.service.ts
DominikDab341: I left it this way to make testing easier.

DominikDab341:
image
image


It seems that this is an issue on the NBP side, not scraper. Yesterday I was able to scrape the currencies normally, but today it stopped working.

DominikDab341:
image

This is output from the same code that was already included in the pull request.

DominikDab341:
I forgot to change the refresh time, I hope it's not a big problem...

src/scrapers/currency-scraper.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  $("table tbody tr").each((_, element) => {
+    const tds = $(element).find("td");
+    const rawCode = $(tds[1]).text().trim();
+    let rawRate = $(tds[2]).text().trim();
+
+    rawRate = rawRate.replace(",", ".");
+    const rate = Number.parseFloat(rawRate);
+
+    if (!Number.isNaN(rate)) {
```
Leadman5555: Failure here is worth logging - it should not fail silently

src/scrapers/currency-scraper.ts
Leadman5555: Bump

src/scrapers/currency-scraper.ts
```diff
@@ -0,0 +1,53 @@
...
+      const code = parts[1];
+
+      const ratePerUnit = rate / multiplier;
+
+      results.push({ code, rate: ratePerUnit });
+    }
+  });
+
+  const wanted = new Set(["USD", "EUR", "GBP"]);
+  const filtered = results.filter((r) => wanted.has(r.code));
```
Leadman5555, line 63: This should also log if no matching values are found. Otherwise we can scrape nothing and nobody will know about it

src/scrapers/currency-scraper.ts
```diff
@@ -0,0 +1,53 @@
...
+  const response = await fetch(
+    "https://nbp.pl/statystyka-i-sprawozdawczosc/kursy/tabela-a",
+  );
+  const html = await response.text();
+
+  const $ = load(html);
+
+  const results: { code: string; rate: number }[] = [];
+
+  $("table tbody tr").each((_, element) => {
```
Leadman5555, line 28: Find the elem first and check for its existence before using it. If it does not exist, the scraper should exit in failure

src/scrapers/currency-scraper.ts
```diff
@@ -0,0 +1,53 @@
...
+      const ratePerUnit = rate / multiplier;
+
+      results.push({ code, rate: ratePerUnit });
+    }
+  });
+
+  const wanted = new Set(["USD", "EUR", "GBP"]);
+  const filtered = results.filter((r) => wanted.has(r.code));
+
+  for (const { code, rate } of filtered) {
```
Leadman5555, line 72: Consider adding a log statement after the loop. How many values were scraped, updated in db, created in db

Leadman5555 (changes requested):
A few minor things to fix in the scraper itself - mainly error handling and logging.

Leadman5555:
> It seems that this is an issue on the NBP side, not scraper. Yesterday I was able to scrape the currencies normally, but today it stopped working.

It's not on the NBP side. You have made far too many requests and got shadow-banned or just simply banned from it looks like. If you want to keep testing the code use a different outgoing ip and/or user agent.

Leadman5555:
For the future, if testing, scrape the site once, save the html, and do not make any more requests - use the saved value instead for tests.

Leadman5555 (approved):
LGTM 👍

PR #243, opened 2025-09-22
Hubert drewlo - task 6 done

Drewienkopl:
Scraping works:
refreshworks

Without currency, just puts PLN and amountPLN is same as amount:
expensesPLN

Conversion works:
conversionWorks


src/currency/currency.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+@Controller("admin/currency")
+export class CurrencyController {
+  constructor(
+    private scraper: RateScraperService,
+    private currency: CurrencyService,
+  ) {}
+
+  @Post("refresh")
+  @Roles("ADMIN")
+  @ApiOperation({ summary: "Create a new trip" })
```
Leadman5555: Remeber to change the description when copy/pasting

src/currency/currency.service.ts
```diff
@@ -0,0 +1,50 @@
+// 👇 Import Prisma-generated model type here
```
Leadman5555: ...

src/currency/currency.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const row = await this.prisma.currencyRate.findUnique({ where: { code } });
+    if (row == null) {
+      throw new NotFoundException(`No rate for currency ${code}`);
+    }
+    return row.rateToPLN;
+  }
+
+  async upsertRate(
+    dtoOrCode: CreateCurrencyRateDto | string,
```
Leadman5555: Split this into two separate methods. Or, leave just the dto here and wrap singular arguments into it

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,70 @@
...
+      "#content .col2.pull-right.module .moduleContent table.ratesTable tbody",
+    );
+
+    if (table.length === 0) {
+      throw new Error("Currency table not found");
+    }
+
+    const currencies = ["USD", "EUR", "GBP"];
+    for (const code of currencies) {
+      const rate = this.extractRate($, table, code);
```
Leadman5555: We are rescanning the whole table for each currency - that's wasteful. Scan the table once and get all needed currencies from a single iteration

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (cells.length >= 3) {
+        const rowName = $(cells[0]).text().trim();
+        if (rowName === currencyName) {
+          const rateString = $(cells[2]).text().trim();
+          rate = Number.parseFloat(rateString);
+        }
+      }
+    });
+
+    if (!rate || Number.isNaN(rate)) {
```
Leadman5555: What would `!rate` mean here? The default value is 0. So it can be either 0, number > 0, or NaN. Change the rate default value to undefined or null

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,70 @@
...
+      throw new TypeError(`Could not parse rate for ${code}`);
```
Leadman5555: If this fails, the whole scraping process will fail. This is not good as we can have at least partial data. Refactor error handling here so failure does not mean instant exit

src/currency/rates-scheduler.service.ts
```diff
@@ -0,0 +1,29 @@
...
+    private currencyService: CurrencyService,
+  ) {}
+
+  @Cron(process.env.RATE_SCRAPE_CRON ?? CronExpression.EVERY_HOUR)
+  async handleCron() {
+    this.logger.log("Starting scheduled rate fetch");
+    try {
+      const rates = await this.scraper.fetchRates();
+      for (const [code, rate] of Object.entries(rates)) {
+        await this.currencyService.upsertRate(code, rate);
```
Leadman5555: Wrap this in a single transaction

src/currency/currency.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+  @ApiCreatedResponse({ description: "Created trip" })
+  @ApiBadRequestResponse({ description: "Invalid input data" })
+  async refresh() {
+    const rates = await this.scraper.fetchRates();
+    for (const [code, rate] of Object.entries(rates)) {
+      await this.currency.upsertRate(code, rate);
```
Leadman5555: Wrap this in a single transaction

Leadman5555 (changes requested):
A few things to fix

Leadman5555:
I haven't seen your review request before (mainly because you committed files after requesting the review and I did not see the review request in the mail box). If I don't respond with review in 24 hours since the request, it's safe to assume I have not noticed it - feel free to ping me on Discord then :+1:

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (table.length === 0) {
+      throw new Error("Currency table not found");
+    }
+
+    table.find("tr").each((_, row) => {
+      const cells = $(row).find("td");
+      if (cells.length >= 3) {
+        const rowName = $(cells[0]).text().trim();
+
+        for (const [code, currencyName] of Object.entries(this.currencyMap)) {
```
Leadman5555: You haven't really changed anything - the complexity is still the same (m*n here, where n is the count of table rows and m is the size of currencyMap). You just flipped the order: instead of for each currency scan the table, you scan the currency map for each table row. Although this solution is probably a fair bit faster as finding an element node in the table takes more time than iterating through an array

Leadman5555 (changes requested):
Well, since this is the last task and we are no longer behind, we can spend more time here 👍 . Refactor the parsing method. The current approach is good, but there is no need to iterate over the `currencyMap` each time.
Hint: allocating a bit more memory here will save us a lot of runtime cost - declare an additional set-like structure. Alternatively, change the logic of the existing map structure if the current mapping direction is sub-optimal

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const table = $(
+      "#content .col2.pull-right.module .moduleContent table.ratesTable tbody",
+    );
+
+    if (table.length === 0) {
+      throw new Error("Currency table not found");
+    }
+
+    table.find("tr").each((_, row) => {
```
Leadman5555: With such lengthy bodies it might be better to just use a simple for loop with `continue` and `break`. Especially if the content requires asynchronous execution (which it does not in this case, so you can leave it if you don't want to change it)

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (Object.keys(rates).length >= expectedCount) {
```
Leadman5555: Tracking the count like this is not really the best idea, due to the `Object.keys()` usage - better to use a separate counter variable

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
...
+        return false;
+      }
+
+      const cells = $(row).find("td");
+      if (cells.length < 3) {
+        return;
```
Leadman5555: This should not be skipped silently - if there are less cells then either the page layout was changed or or code did not work correctly

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      const rowName = $(cells[0]).text().trim();
+      const code = this.nameToCode.get(rowName);
+
+      if (code == null) {
+        return;
+      }
+
+      if (Object.hasOwn(rates, code)) {
```
Leadman5555: Do not use the generic object methods if more specific and optimal alternatives exist

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
...
+        return;
```
Leadman5555: If a double rate for the same currency is listed, than such behavior should be at the very least logged.

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    for (const wantedCode of Object.keys(this.currencyMap)) {
+      if (!(wantedCode in rates)) {
+        this.logger.warn(`Rate not found for ${wantedCode}`);
+      }
+    }
+
+    if (Object.keys(rates).length === 0) {
+      this.logger.error("No valid currency rates were parsed");
```
Leadman5555: This should throw - no rates equals scrapping failed

src/currency/rate-scraper.service.ts
```diff
@@ -0,0 +1,90 @@
+import * as cheerio from "cheerio";
+
+import { Injectable, Logger } from "@nestjs/common";
+
+@Injectable()
+export class RateScraperService {
+  private readonly logger = new Logger(RateScraperService.name);
+
+  private readonly currencyMap: Record<string, string> = {
```
Leadman5555: If this  a constant that we are not modifying in any way, then place is as a const instead of a class field. That way other modules would also be able to see what currencies are available.

Leadman5555 (changes requested):
Probably the last round of nitpick changes. Although the review is even more detailed in a normal project.

Leadman5555 (approved):
Alright, let's wrap it up :+1:

